PR #7, opened 2016-07-21
allMarkers file to replace existing marker files for civ.cop.med.admin

fresqo:
Quickly went through the code and made some adjustments. I tested this locally on my own, worked fine for my own marker. Needs to be tested multiplayer + vehicles etc. Also little concerned of how a crew of people would look as i left the code as it was but it looks a little funny to me.. also. any reasons why we are using (for ""_i"" from 0 to 1 step 0 do ) instead of like a while statement?

### DIFF
--- a/life_server/functions.sqf
+++ b/life_server/functions.sqf
@@ -309,6 +309,7 @@ compileFinal "
 publicVariable "TON_fnc_clientMessage";
 
 TON_fnc_MapMarkersAdmin = compileFinal "
+  private[""_color""];
   life_markers_Vehicles = [];
   life_markers_Players = [];
   if (!life_markers) then {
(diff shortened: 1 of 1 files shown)


BoGuu:
Couple things:

I think this needs entirely rewriting, as it's 100% pure ugliness currently. This can also be moved into the clientside files under \admin\, it doesn't need to be PVAR'd.

The major issue we have currently is the fact none of the marker loops are closed when the map is closed, so they continue running, and thus stack when you open it again. We can use the new eventhandler (Map : https://community.bistudio.com/wiki/Arma_3:_Event_Handlers/addMissionEventHandler#Map) to open and close these now, as it's just been added.

fresqo:
Oh thank you, I thought it was odd too, I will re write it

fresqo:
ok so.. i have gotten rid of all the adminMarkers,civMarkers,medicMarkers,copMarkers files and made an Allmarkers file to do all of them. Using the addmissioneventhandler also.. atm i have vehicles and players showing with admin markers but im thinking of making a selection when clicked to pick either vehicles or players?

Jawshy:
Can I delete this branch (fresqo-patch-1)?

fresqo:
Yes please

PR #42, opened 2016-07-30
EscapeMenu Updates

tkcjesse:
I'm back bitches. 

Here's a preview. https://gyazo.com/2cba6f99d6263eb9faadfcf2b16232c7

The UID is mandatory. Obviously you can change it if ya really want. Didn't want to make a ton of options for something so simple...

**Why?** Glad you asked. Because BI has removed showing players their UID in their profile. There is no "easy" way to now see it AFAIK. This is an easy way for server owners to be able to still ask players to provide their UID for admin support. Was going to add something on the Y Menu, thought this would be the best way, as well as offering a way for some customization.


tkcjesse:
@BoGuu Lets go. Too lazy to make branches.

Jawshy:
> [...] Because BI has removed showing players their UID in their profile. [...]

Try going to **profile** and under **unit** click **edit**. Change **unit** from **none** to **custom**. A player ID field should appear.

tkcjesse:
Thats stupid... lol. Still I think for newbie players and such this will be better. If you're against it do say so. However i feel eliminating the grayed out field manual and save option was much needed.

Jawshy:
I'm fine with this. The ability to [copy the UID to the clipboard](https://community.bistudio.com/wiki/copyToClipboard) would be even more useful, but that command is disabled in multiplayer.

tkcjesse:
Yea. Didn't mean it rude or anything. I'm open to ideas

PR #49, opened 2016-08-01
format Consistency

tkcjesse:


Jawshy:
I didn't find any obvious errors when testing this; not that there should be any.

PR #76, opened 2016-08-25
Update fn_boltcutter.sqf

fresqo:
Resolves #<issue ID here>. 

- there was no timer for the Tanoa vault door

### DIFF
--- a/Altis_Life.Altis/core/items/fn_boltcutter.sqf
+++ b/Altis_Life.Altis/core/items/fn_boltcutter.sqf
@@ -53,7 +53,7 @@ _cP = 0.01;
 
 switch (typeOf _building) do {
     case "Land_Dome_Big_F": {_cpRate = 0.003;};
-    case "Land_Medevac_house_V1_F";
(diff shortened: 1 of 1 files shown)


BoGuu:
Have you actually verified there is no timer? I made this on purpose, it should be ```_cpRate = 0.0015```

See https://en.wikipedia.org/wiki/Switch_statement#Fallthrough

fresqo:
oh i thought it was an error will test when i get home.
but wouldnt it have to look like this...
> switch (typeOf _building) do {
    case "Land_Dome_Big_F": {_cpRate = 0.003;};
    case "Land_Medevac_house_V1_F": 
    case "Land_Research_house_V1_F": {_cpRate = 0.0015;};
    break;
    default {_cpRate = 0.08;}
};

BoGuu:
Depends how many errors you like! ;) You don't want it to start executing at that case, which the semi colon would allow it to.

fresqo:
ok checked your way works sorry .. my mistake..

PR #88, opened 2016-09-10
ActionKeyHandler Fixes

harmdhast:
#### **Changes proposed in this pull request:**

> • Allow cops to use the ATMs with the action key (Windows or custom user key 10)
> • Side check for cop menu (Fix for civilian admin escort on some admin menus)
> • I don't know why it was handled that way tbh

### DIFF
--- a/Altis_Life.Altis/core/functions/fn_actionKeyHandler.sqf
+++ b/Altis_Life.Altis/core/functions/fn_actionKeyHandler.sqf
@@ -13,15 +13,15 @@ if (life_action_inUse) exitWith {}; //Action is in use, exit to prevent spamming
 if (life_interrupted) exitWith {life_interrupted = false;};
 _isWater = surfaceIsWater (visiblePositionASL player);
 
+if (playerSide isEqualTo west && player getVariable ["isEscorting",false]) exitWith {
+    [] call life_fnc_copInteractionMenu;
+};
+
 if (LIFE_SETTINGS(getNumber,"global_ATM") isEqualTo 1) then{
     //Check if the player is near an ATM.
     if ((call life_fnc_nearATM) && {!dialog}) exitWith {
         [] call life_fnc_atmMenu;
     };
-};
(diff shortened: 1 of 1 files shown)


BoGuu:
Calling the cop interaction menu for any escorters is indeed bad design, but obviously not an issue with the code base (just external as you said). Do you think ATMs should take precedence over the interaction menu?

harmdhast:
In my dream world we would get rid of the addActions. And allowing cops to use the same ATMs as the civilians seems logic to me. Server admins tends to forget to add the ATM code in the editor-placed ones and you can't access map / world placed ATM without the cursorTarget / cursorObject. (You still can get the ATM positions from the cfgWorld then use nearestObject to add an action but yeah, i don't think everyone can do that).
That's just my thoughts anyway.

BoGuu:
>And allowing cops to use the same ATMs as the civilians seems logic to me
This code is just stopping them using the ATM while escorting someone, what do you mean exactly?

Currently any editor placed ATM's will be accessible via the action key without an addAction.

harmdhast:
I don't think you understand
```
player getVariable ["isEscorting",false]; // When cop is not escorting then trigger the menu
```
I doesn't seems logical to me or I missunderstood. Tell me.

BoGuu:
`player getVariable ["isEscorting", false]`

Will return true if the person is escorting someone. The false is purely there to avoid undefined errors. Take a look at the getVariable wiki page if you're confused

harmdhast:
Wow my brain melted beetween setVariable and getVariable.
There's still the side check tbh.

harmdhast:
Btw BoGuu is there any way I can contact you personnaly ? (Discord / Skype ?)

BoGuu:
@harmdhast

I'm currently on holiday, but that looks better. Could you add lazy eval to it before merge?

You can contact me on discord. There is a link to a server I'm in on the README

Issue #96, opened 2016-09-21
Wrong param expected type for saving civ_alive


macchky:
### Expected behaviour
When you rejoin the server as civ, you spawn at last location.

### Actual behaviour
Respawn dialog showed up.

### Steps to reproduce the behaviour 
1. just abort

### Probrem code

> life_server\Functions\MySQL\fn_updateRequest.sqf at line 17

```
_alive = [_this,9,0,[0]] call BIS_fnc_param;
_position = [_this,10,[],[[]]] call BIS_fnc_param;
```
this [_this,9,**0,[0]**] is wrong statement. 
_alive passed as true or false , but that expects data as 0 or 1 , so _alive always become 0......

----

Game version: 
1.62
Branch: 
main

Jawshy:
@macchky, what are your Config_Master.hpp settings? [Specifically these](https://github.com/AsYetUntitled/Framework/blob/master/Altis_Life.Altis/config/Config_Master.hpp#L24-L27).

macchky:
Propery configured.

```
    save_civilian_weapons = true; //Allow civilians to save weapons on them?
    save_civilian_position = true; //Save civilian location?
    save_civilian_position_restart = true; //Save civilian location only between restarts. After a server restart you'll have to spawn again.
    /* !!!TO SAVE POSITION BETWEEN RESTARTS save_civilian_position MUST BE TRUE!!! */
    save_civilian_positionStrict = true;
```

Anyway life_server\Functions\MySQL\fn_updateRequest.sqf at line 17 make civ_alive to 0 by `[] call SOCK_fnc_updateRequest`

Jawshy:
Feel free to submit a pull request if you would like to do so.

Jawshy:
Hmm. I cannot reproduce this issue on the [current build](https://github.com/AsYetUntitled/Framework/commit/7ca0b5b6ec42f223f27fff185e1d841a702e7c1c). That is to say that `save_civilian_position` is working as intended.

BoGuu:
@Jawshy 

Try sync then combat log

Jawshy:
Nah-dah. Kicked myself via `#kick` and RCon; spawned at last saved positions.

```sql
civ_alive
1
```

PR #97, opened 2016-09-21
#96 Fix wrong param expect data for civ_alive

macchky:
#96


setoy:
Aren't you missing out the HC?

macchky:
Oh. Re pull request

macchky:
Added

BoGuu:
@macchky 
Sorry this is delayed, have been incredibly busy.

You are indeed right about the expected type being wrong, however, this surprisingly has no implication.

The smart cookie who added the position saving into the build decided not to use the variable defined, instead using the raw data from the variables passed.

PR #133, opened 2016-10-16
Whitelisted Terrain Sorting Function (SpyGlass)

SimZor:

PR #135, opened 2016-10-17
Remove serv_wanted_remove from SpyGlass

SimZor:
Removes serv_wanted_remove from SpyGlass config which is not used anymore


Jawshy:
Would have been better to have gone with PR https://github.com/AsYetUntitled/Framework/pull/133 [Whitelisted Terrain Sorting Function (SpyGlass)].

SimZor:
Removed variable in this PR https://github.com/AsYetUntitled/Framework/pull/133.

Closing.

PR #149, opened 2016-10-27
Update

Jason2605:
- Change some old config to the new config


setoy:
May I put my change requests to the config system here? If so:

- The [descriptions](https://github.com/AsYetUntitled/Framework/blob/c0590e5c9c51b52f28be8132714a595c05e6b162/Altis_Life.Altis/config/Config_Clothing.hpp#L6-L9) haven't been updated in all configs.
- And isn't Config_vItems.hpp missing?
- And why is it condition_ in Config_Clothing.hpp, but condition**s** in all others?
- And what about a "master" switch in Config_Vehicles.hpp to restrict a vehicle shop? Instead it's now specified in each and every vehicle. Like in Config_Clothing.hpp.
- No spawns are shown (except own houses) even when `conditions = ""`

Jason2605:
Alright, ill have a look. What do you mean Config_vItems is missing? And with the spawns can you upload your client RPT please.

setoy:
> What do you mean Config_vItems is missing?

Config_vItems.hpp is not using the "new" `conditions` statement, but the "old" level and license checks.

No errors in client RPT, but I'll try to fetch one later.

Jason2605:
https://github.com/AsYetUntitled/Framework/blob/master/Altis_Life.Altis/config/Config_vItems.hpp#L14

Is it? Also I've just downloaded the mission and the spawn points worked perfectly for me.

Jason2605:
@setoy

setoy:
@Jason2605 Went to sleep, I'll check. Thx.

setoy:
Here's the corresponding [client RPT](http://pastebin.com/8SVpD4GD). The error is most likely due to the spawn points array being empty.

Jason2605:
Should be fixed in the latest push @setoy

setoy:
> Should be fixed in the latest push @setoy

Confirmed working.

Jason2605:
https://github.com/AsYetUntitled/Framework/blob/8ec018c84a32569b863c24d05e991578c9f2617d/ApexContent/Config_Vehicles.hpp#L636

setoy:
https://github.com/AsYetUntitled/Framework/blob/74032cff258be23328c88960a75fe3e923a0ff1a/Altis_Life.Altis/core/shops/fn_vehicleShopBuy.sqf#L54

It's `STR_Shop_Veh_NoLicense`

Jason2605:
@setoy The NoLicense string was modified to say license/level, and D: merge conflict

setoy:
@Jason2605 Yeah, but why not use `STR_Shop_Veh_NotAllowed` which just says "No access"... Should be enough :p

Jason2605:
@setoy You just posted the same string twice? And the hint it shows is really personal preference, as it shows they need a license/level instead of just 'Not allowed'. If you really don't like it, you can change it when you use it.

setoy:
> @setoy Ah right typo probably.

That's what I wanted to say :D

PR #151, opened 2016-10-29
Client only briefing

trinitynet:
this should not be ran on the server and just causes a perm locked
thread as it waits for a display which will never show


Altis_Life.Altis/init.sqf
```diff
@@ -7,7 +7,9 @@
 */
 StartProgress = false;
 
-[] execVM "briefing.sqf"; //Load Briefing
+if (hasInterface) then {
```
DarkSilencerZA: Some peoples place more than this in their inits. Would it not be better to use this:

`if (!hasInterface) exitWith {
	StartProgress = true;
};`

trinitynet: eh

PR #152, opened 2016-10-29
Client only briefing

trinitynet:
this should not be ran on the server and just causes a perm locked
thread as it waits for a display which will never show


trinitynet:
ignore

PR #162, opened 2016-11-04
Better way of using a switch...

luudvkeulen:


BoGuu:
A couple things. Please add more body to your pulls. Secondly, as correct as this small change is, it really doesn't affect performance whatsoever. So if you do want to make pulls like this, perhaps finding other uses/tweaks as well, versus just one.

luudvkeulen:
Allright, I am editing the framework for my own server but I do not have the whole repo pulled so when I find something I just edit it in github directly hence why I only PR one file at a time.

BoGuu:
LGTM

PR #169, opened 2016-11-12
Fix incorrect variable usage in WantedSystem

OfficialNoms:
Fix incorrect variable usage in WantedSystem (_crime instead of _crimes)


BoGuu:
LGTM

PR #178, opened 2016-11-17
Update stringtable.xml for added french translation

Lokiro67:
- Add translated text in french


Lokiro67:
ok no prob i'm working in local for updated later after approved ^^

PR #182, opened 2016-11-20
disableChannels Issue Workaround

Jawshy:
- Fixes issue with [`disableChannels`](https://github.com/AsYetUntitled/Framework/blob/a3a7cc9626b6dabc335c2677d1588bbc0d10d268/Altis_Life.Altis/description.ext#L1) command not working if a client aborts to the lobby and then rejoins the mission. 
  - Remove if and when Bohemia actually fixes the issue. 
  - Issue ticket: [disableChannels settings cease to work when leaving/rejoining mission](https://feedback.bistudio.com/T117205)

### DIFF
--- a/Altis_Life.Altis/core/init.sqf
+++ b/Altis_Life.Altis/core/init.sqf
@@ -1,7 +1,7 @@
 #include "..\script_macros.hpp"
 /*
     File: init.sqf
-    Author:
+    Author: Bryan "Tonic" Boardwine
 
     Description:
     Master client initialization file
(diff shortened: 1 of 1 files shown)


setoy:
You're accessing the missionConfigFile, yet there's no corresponding setting inside the config? Is that right?

Jawshy:
>You're accessing the missionConfigFile, yet there's no corresponding setting inside the config? Is that right?

It is referring to [this](https://github.com/AsYetUntitled/Framework/blob/a3a7cc9626b6dabc335c2677d1588bbc0d10d268/Altis_Life.Altis/description.ext#L1), @setoy.

PR #183, opened 2016-11-21
Update ticketing private

SimZor:
Resolves No Issue

- Private being an array, it doesn't need to be.

### DIFF
--- a/Altis_Life.Altis/core/cop/fn_ticketGive.sqf
+++ b/Altis_Life.Altis/core/cop/fn_ticketGive.sqf
@@ -6,7 +6,7 @@
     Description:
     Gives a ticket to the targeted player.
 */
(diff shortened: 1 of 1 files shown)


Altis_Life.Altis/core/cop/fn_ticketGive.sqf
```diff
@@ -6,7 +6,7 @@
     Description:
     Gives a ticket to the targeted player.
 */
-private ["_val"];
+private "_val";
```
BoGuu: Or by the variable itself ;)?

SimZor: yeah didnt think that far 👍

BoGuu:
LGTM

PR #241, opened 2017-01-13
db: Update to extDB3

danielstuart14:
Next arma update will support 64-bit and as extDB2 can't be build in x64
(without changing its code), it's about time to support extDB3.
Also, extDB3 already supports linux and doesn't have any major issue.

I know there're plans for using sql-custom, but, we can migrate to sql-custom later. 
extDB3 is a priority as we won't be able to use x64 arma servers with extDB2.

- Updated extDB version support.


BoGuu:
Minus the params (plan was to remove all MRES entirely, as it's not needed): https://github.com/GetSomePanda/AL-Custom-extDB3/blob/master/%40extDB3/sql_custom/altis-life-custom.ini

danielstuart14:
we could just merge extDB3 and then, merge all the changes after that.

Here is one more usefull feature of extDB3: 
```
"extDB3" callExtension "9:UTC_TIME"
```
makes possible to get rid of BEC (since it hasn't been update for a long) for restart messages. (with real time warnings)

BoGuu:
> Here is one more usefull feature of extDB3

extDB2 has that as well :)

LGTM @Jawshy

 I'll add predefineds soon

trinitynet:
tested and works from what i can see

life_hc/hc_macros.hpp
```diff
@@ -35,7 +35,7 @@
 #define CONFIG_BOOL(NUMBER) [##NUMBER] call { _ret = false; if ((_this select 0) in [0,1] && (_this select 0) isEqualTo 1) then { _ret = true; }; _ret;}
 
 // extDB2 Macros
```
Jawshy: `extDB2` :japanese_goblin:

Jawshy:
LGTM

PR #242, opened 2017-01-14
config: spy: Add missing functions & clean

danielstuart14:
Spy. Config was missing spyglass functions.
Also, break the lines so it's better for comparing changes in github.

- Add missing functions after this commit: 1ea63ea409ad7bb058234677607ae5035f74fd37
- Make it better to compare changes.
- Only check if the UI var is in the UI array.


trinitynet:
yeh i forgot to add that to the PR. my bad

PR #247, opened 2017-01-18
functions: move from server to client side.

danielstuart14:
"Using publicVariable too frequently in a given period of time can cause other parts of the game to experience bandwidth problems."

- Move some of the functions to client-side.
- Remove KRON_Strings
- isNumber: Switch from Kron to official splitString command.
- getDPMission: Switch from Kron to official splitString | joinString command.
- getIndex: Optimized the code
- ATM: Cleaned the code, fixed an exploit in gangWithdraw &  added maxvalue_ATM config.
- numberText: Localized number separator. (Most european countries use space instead comma)


coresync2k:
What are the benefits of the changes?

danielstuart14:
```
 Using publicVariable too frequently in a given period of time can cause other parts of the game to experience bandwidth problems.
```
from: https://community.bistudio.com/wiki/publicVariable

If you don't need to keep broadcasting a var for each player that joins, why do that?

Jawshy:
> **There isn't why set a function** as publicVariable if we can just move it to client side.

Could you rephrase that?

SimZor:
So no more KRON_Strings with the new number function?

danielstuart14:
There're some scripts that still using KRON_Strings.
That's why it isn't ready for merge.

I'll be uploading the changes here.

Altis_Life.Altis/core/functions/fn_isNumber.sqf
```diff
@@ -0,0 +1,27 @@
...
+    Parameter(s):
+        0: STRING - String to be analyzed
+
+    Returns:
+    BOOL
+*/
+params [
+    ["_array","",[""]]
+];
+_array = _array splitString "";
```
BoGuu: We could use toArray and replace the strings below w/ just numbers instead, seems more logical. Can also use a count

Altis_Life.Altis/core/functions/fn_getIndex.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+_return = -1;
+{
+    if ((_x isEqualType [] && {_value in _x}) || {_value isEqualTo _x}) then {
+        _return = _forEachIndex;
+        breakOut "";
+    };
+} forEach _array;
+
+_return;
```
BoGuu: I'm 50/50 on this due to the fact the user could avoid using this (and use the other commands that do the same thing, internally). I know it's good to be 'well-rounded', but I think most people know when and when to not use this. It would seem silly to use this vs find. I like the idea of another function supporting a deeper search though (actual index path) on a side note. If we do stick with this, perhaps abstracting the logic out of the loop, so for large arrays, it doesn't need to run un-needed conditions every iteration?

BoGuu:
> If you don't need to keep broadcasting a var for each player that joins, why do that?

Well it's the JIP queue, but either way, as I'm sure we've talked about many times in the past, the functions server side like this are just :/

BoGuu:
As for https://github.com/AsYetUntitled/Framework/blob/master/Altis_Life.Altis/core/actions/fn_getDPMission.sqf#L26

There is a 1000 things we could do, even as far as making a function that'd replace characters in string (if it'd ever come in handy?), if not, something inline. Using a string replacement function was silly for only replacing chars anyway

Altis_Life.Altis/core/functions/fn_isNumber.sqf
danielstuart14: If we're going this way, we can get rid of this function and use parseNumber + round.

Like:
parseNumber (_string);
round _string;

This way we get rid of both letters and symbols.

what do you think?

danielstuart14:
ready.

https://github.com/danielstuart14/Framework/blob/d3f9ef93783ddeeb81284e686a4520739ee46d2b/Altis_Life.Altis/core/functions/fn_isNumber.sqf#L22 
After doing some tests, seems like using greater|less than is faster than checking if it's in an array. (A little faster if it has only numbers, much faster if it hasn't.)

https://github.com/danielstuart14/Framework/blob/d3f9ef93783ddeeb81284e686a4520739ee46d2b/Altis_Life.Altis/core/functions/fn_numberText.sqf#L27
From here: http://www.statisticalconsultants.co.nz/blog/how-the-world-separates-its-digits.html

@BoGuu

danielstuart14:
https://github.com/AsYetUntitled/Framework/pull/247/commits/14385c7413506a7e684d30d5a88502144f007774#diff-332f9c8b0914442c43820d830d064f8aL13
There isn't why set a max value for compensate. Also, I already removed the string in the earlier commit.

danielstuart14:
tested, should be okay.

danielstuart14:
@BoGuu should something be changed?

Altis_Life.Altis/core/actions/fn_getDPMission.sqf
```diff
@@ -23,7 +27,7 @@ life_dp_start = _target;
 life_delivery_in_progress = true;
 life_dp_point = call compile format ["%1",_dp];
 
-_dp = [_dp,"_"," "] call KRON_Replace;
+_dp = (_dp splitString "_") joinString " ";
```
BoGuu: Can just add the private directly here

Altis_Life.Altis/core/admin/fn_adminCompensate.sqf
```diff
@@ -8,9 +8,10 @@
 */
 private ["_value","_action"];
 if (FETCH_CONST(life_adminlevel) < 2) exitWith {closeDialog 0; hint localize "STR_ANOTF_ErrorLevel";};
-_value = parseNumber(ctrlText 9922);
-if (_value < 0) exitWith {};
-if (_value > 999999) exitWith {hint localize "STR_ANOTF_Fail"};
+_value = ctrlText 9922;
+if (!([_value] call life_fnc_isNumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+_value = parseNumber(_value);
+if (_value < 1) exitWith {};
```
BoGuu: I think you should just parseNumber the input, then check if it's 0 (no number parsed, or they just put 0, either way, doesn't matter)

Altis_Life.Altis/core/functions/fn_getIndex.sqf
BoGuu: This looks good now, but would be better to define scopeName "xxx" outside the loop, then use `_return breakOut "xxx", instead of just breaking out the loop (which is just an exitWith), a pseudo C like return

Altis_Life.Altis/core/functions/fn_isNumber.sqf
```diff
@@ -0,0 +1,28 @@
...
+];
+_array = toArray _array;
+
+_return = true;
+{
+    if (_x < 48 || _x > 57) then {
+        _return = false;
+        breakOut "";
+    };
+} count _array;
```
BoGuu: Same comment with breakOut here

Altis_Life.Altis/core/functions/fn_numberText.sqf
```diff
@@ -1,28 +1,35 @@
...
-_modBase = _digitsCount % _mod;
+_modBase = _digitsCount % 3;
 _numberText = "";
+
+_mark = ",";
+if (!(language isEqualTo "English")) then { _mark = " "; };
 {
     _numberText = _numberText + str _x;
-    if ((_foreachindex - _modBase) % (_mod) isEqualTo 0 && !(_foreachindex isEqualTo _digitsCount)) then {_numberText = _numberText + ",";};
+    if (((_foreachindex - _modBase) % 3) isEqualTo 0 && !(_foreachindex isEqualTo _digitsCount)) then {
```
BoGuu: Can lazy eval this

Altis_Life.Altis/dialog/function/fn_bankDeposit.sqf
```diff
@@ -6,13 +6,14 @@
     Description:
     Figure it out.
 */
-private ["_value"];
-_value = parseNumber(ctrlText 2702);
+private "_value";
+_value = ctrlText 2702;
```
BoGuu: Can move this private down to _value

Altis_Life.Altis/dialog/function/fn_bankDeposit.sqf
```diff
@@ -6,13 +6,14 @@
...
+if (!([_value] call life_fnc_isNumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+_value = parseNumber(_value);
```
BoGuu: Same idea as before, just parse it and check <= 0

Altis_Life.Altis/dialog/function/fn_bankDeposit.sqf
```diff
@@ -6,13 +6,14 @@
...
 
 //Series of stupid checks
-if (_value > 999999) exitWith {hint localize "STR_ATM_GreaterThan";};
-if (_value < 0) exitWith {};
-if (!([str(_value)] call TON_fnc_isnumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+if (_value > (LIFE_SETTINGS(getNumber,"maxvalue_ATM"))) exitWith {hint format [localize "STR_ATM_GreaterThan",[LIFE_SETTINGS(getNumber,"maxvalue_ATM")] call life_fnc_numberText];};
```
BoGuu: Can define a variable for the config value here, so we don't fetch it twice

Altis_Life.Altis/dialog/function/fn_bankTransfer.sqf
```diff
@@ -7,20 +7,21 @@
     Figure it out again.
 */
 private ["_value","_unit","_tax"];
-_value = parseNumber(ctrlText 2702);
-_unit = call compile format ["%1",(lbData[2703,(lbCurSel 2703)])];
-if (isNull _unit) exitWith {};
+_value = ctrlText 2702;
+if (!([_value] call life_fnc_isNumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+_value = parseNumber(_value);
```
BoGuu: Same parse idea

Altis_Life.Altis/dialog/function/fn_bankWithdraw.sqf
```diff
@@ -6,16 +6,20 @@
...
 */
-private ["_value"];
-_value = parseNumber(ctrlText 2702);
-if (_value > 999999) exitWith {hint localize "STR_ATM_WithdrawMax";};
-if (_value < 0) exitWith {};
-if (!([str(_value)] call TON_fnc_isnumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+private "_value";
+_value = ctrlText 2702;
+if (!([_value] call life_fnc_isNumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+_value = parseNumber(_value);
```
BoGuu: Ditto

Altis_Life.Altis/dialog/function/fn_bankWithdraw.sqf
```diff
@@ -6,16 +6,20 @@
...
+
+//Series of stupid checks
+if (_value > (LIFE_SETTINGS(getNumber,"maxvalue_ATM"))) exitWith {hint format [localize "STR_ATM_WithdrawMax",[LIFE_SETTINGS(getNumber,"maxvalue_ATM")] call life_fnc_numberText];};
```
BoGuu: Ditto

Altis_Life.Altis/dialog/function/fn_gangDeposit.sqf
```diff
@@ -6,26 +6,26 @@
     Description:
     Deposits money into the players gang bank.
 */
-private ["_value"];
-_value = parseNumber(ctrlText 2702);
+private ["_value","_gFund"];
+_value = ctrlText 2702;
+if (!([_value] call life_fnc_isNumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+_value = parseNumber(_value);
```
BoGuu: Ditto

Altis_Life.Altis/dialog/function/fn_gangDeposit.sqf
```diff
@@ -6,26 +6,26 @@
...
 group player setVariable ["gbank_in_use_by",player,true];
 
 //Series of stupid checks
-if (_value > 999999) exitWith {hint localize "STR_ATM_GreaterThan";};
-if (_value < 0) exitWith {};
-if (!([str(_value)] call TON_fnc_isnumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+if (_value > (LIFE_SETTINGS(getNumber,"maxvalue_ATM"))) exitWith {hint format [localize "STR_ATM_GreaterThan",[LIFE_SETTINGS(getNumber,"maxvalue_ATM")] call life_fnc_numberText];};
+if (_value < 1) exitWith {};
```
BoGuu: Ditto with config

Altis_Life.Altis/dialog/function/fn_gangWithdraw.sqf
```diff
@@ -6,22 +6,22 @@
     Description:
     Withdraws money from the gang bank.
 */
-private ["_value"];
-_value = parseNumber(ctrlText 2702);
-_gFund = GANG_FUNDS;
+private ["_value","_gFund"];
+_value = ctrlText 2702;
+if (!([_value] call life_fnc_isNumber)) exitWith {hint localize "STR_ATM_notnumeric"};
+_value = parseNumber(_value);
```
BoGuu: Ditto :D

Altis_Life.Altis/dialog/function/fn_safeStore.sqf
```diff
@@ -12,7 +12,7 @@ _ctrl = CONTROL_DATA(3503);
 _num = ctrlText 3506;
 
 //Error checks
-if (!([_num] call TON_fnc_isnumber)) exitWith {hint localize "STR_MISC_WrongNumFormat";};
+if (!([_num] call life_fnc_isNumber)) exitWith {hint localize "STR_MISC_WrongNumFormat";};
```
BoGuu: Ditto

Altis_Life.Altis/dialog/function/fn_safeTake.sqf
```diff
@@ -15,7 +15,7 @@ _num = ctrlText 3505;
 _safeInfo = life_safeObj getVariable ["safe",0];
 
 //Error checks
-if (!([_num] call TON_fnc_isnumber)) exitWith {hint localize "STR_MISC_WrongNumFormat";};
+if (!([_num] call life_fnc_isNumber)) exitWith {hint localize "STR_MISC_WrongNumFormat";};
```
BoGuu: Ditto

BoGuu:
@danielstuart14 

Sorry for the late reply, has been a busy few days. I marked everything I've seen, sorry that it's a lot, should make it easier though :P

danielstuart14:
@BoGuu thx, I'll be doing the needed changes. :D

coresync2k:
https://github.com/AsYetUntitled/Framework/blob/master/life_server/Functions/WantedSystem/fn_wantedAdd.sqf#L22 

must be changed.

BoGuu:
Any etc @danielstuart14 ?

danielstuart14:
Sorry about being afk, college took all my time in the last months. 
I'm going to do all the changes this week.

Altis_Life.Altis/core/admin/fn_adminCompensate.sqf
danielstuart14: We can't do that, because for example:
45A5B2C will result a number greater than 0.
Also, isNumber prevents fractions. (Ex: 4502.5)

DomT602:
As fork has been deleted, and with recent sweeping changes, this PR will be closed, however functions.sqf will be moved into appropriate place for 6.x.

PR #254, opened 2017-01-23
Missing variables on death

J4Wx:
On death, life_action_inUse is not declared false, it is unlikely that this will matter, but is bound to occur at some point, and since this would make users unable to perform certain actions, this would be game breaking until reconnecting in those cases.

- Declare this variable as false on Player Killed, on Revive and on Respawn.

_Sorry about onPlayerKilled... couldn't get github to recognise that I hadn't changed anything on 113-115._


DarkSilencerZA:
This should only need to be called on the player being killed. The other two scripts run after.

BoGuu:
Not sure this will ever be an issue, but no harm done in being cautious

PR #258, opened 2017-01-26
Including "CfgRemoteExec.hpp" to life_server/config.cpp

### LINKED ISSUE #257: global remote execution not allowed in life_server
### Expected behaviour
A code like `[_vehicle, ["RopeAttach", {ropeCut [ _this select 1, 1];}]] remoteExec ["addEventHandler",0,true];
` should be executed in Functions/Systems/fn_spawnVehicle.

### Actual behaviour
It doesn't as JIP doesn't seem to be enabled or whatever.

### [RPT]
NO RPT

----

Mission version: Current

Launcher version: Current

Game version: Current

Branch: Altis Life 5.0

Blauschimmel:

Resolves #257 

Including "CfgRemoteExec.hpp" to the config.cpp in life_server

### DIFF
--- a/life_server/config.cpp
+++ b/life_server/config.cpp
@@ -129,3 +129,5 @@ class CfgVehicles {
         };
     };
 };
+
+#include "CfgRemoteExec.hpp"
--- a/life_server/config.cpp
+++ b/life_server/config.cpp
@@ -130,4 +130,4 @@ class CfgVehicles {
     };
 };
 
-#include "CfgRemoteExec.hpp"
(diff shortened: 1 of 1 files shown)


Blauschimmel:
I don't know where you guys prefer to include other files, so I just put it in the end of file...

BoGuu:
Was this an accident ;)?

Blauschimmel:
Nope I just realized that it also works without including it. I'm really confused, so I'm going to check that first. Because it used to not work before...

PR #269, opened 2017-02-05
Removed query char limit

harmdhast:
The usage of format limits queries and answers to ~8000 chars.
Fixed it with formatText.
New limit is ~1000000 chars.
(You gotta do something for the database part one day)


harmdhast:
Needing more testing for now.
(Weird behavior)

PR #270, opened 2017-02-05
Remove query char limit v2

harmdhast:
The usage of format limits queries and answers to ~8000 chars.
Fixed it with formatText + str convert for compile and callExtension
New limit is ~1000000 chars.
(You gotta do something for the database part one day)

### DIFF
--- a/life_hc/MySQL/General/fn_asyncCall.sqf
+++ b/life_hc/MySQL/General/fn_asyncCall.sqf
@@ -16,19 +16,19 @@ _queryStmt = [_this,0,"",[""]] call BIS_fnc_param;
 _mode = [_this,1,1,[0]] call BIS_fnc_param;
 _multiarr = [_this,2,false,[false]] call BIS_fnc_param;
 
-_key = EXTDB formatText ["%1:%2:%3",_mode,FETCH_CONST(life_sql_id),_queryStmt];
+_key = EXTDB str formatText ["%1:%2:%3",_mode,FETCH_CONST(life_sql_id),_queryStmt];
 
 if (_mode isEqualTo 1) exitWith {true};
 
 _key = call compile formatText ["%1",_key];
 _key = (_key select 1);
-_queryResult = EXTDB formatText ["4:%1", _key];
+_queryResult = EXTDB str formatText ["4:%1", _key];
 
 //Make sure the data is received
 if (_queryResult isEqualTo "[3]") then {
     for "_i" from 0 to 1 step 0 do {
         if (!(_queryResult isEqualTo "[3]")) exitWith {};
-        _queryResult = EXTDB formatText ["4:%1", _key];
+        _queryResult = EXTDB str formatText ["4:%1", _key];
     };
 };
 
@@ -37,7 +37,7 @@ if (_queryResult isEqualTo "[5]") then {
     for "_i" from 0 to 1 step 0 do { // extDB3 returned that result is Multi-Part Message
         _queryResult = "";
         for "_i" from 0 to 1 step 0 do {
-            _pipe = EXTDB formatText ["5:%1", _key];
+            _pipe = EXTDB str formatText ["5:%1", _key];
             if (_pipe isEqualTo "") exitWith {_loop = false};
             _queryResult = _queryResult + _pipe;
         };
--- a/life_hc/MySQL/General/fn_asyncCall.sqf
+++ b/life_hc/MySQL/General/fn_asyncCall.sqf
@@ -20,7 +20,7 @@ _key = EXTDB str formatText ["%1:%2:%3",_mode,FETCH_CONST(life_sql_id),_queryStm
 
 if (_mode isEqualTo 1) exitWith {true};
 
-_key = call compile formatText ["%1",_key];
+_key = call compile str formatText ["%1",_key];
 _key = (_key select 1);
 _queryResult = EXTDB str formatText ["4:%1", _key];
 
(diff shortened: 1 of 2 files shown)


BoGuu:
I'm not sure I quite understand what you think is happening, the result (which could be 10k chars indeed) is never formatted on return. The only issue could be that your query itself is over 8.2k chars *or* the ticket ID extDB returns is (which is physically impossible in this case).

harmdhast:
Indeed.
Still you can't send 8k+ queries.
I feel we should not be limited in chars. 
Anyway I did it on my system (I use a huge fucking amount of join, update case... which takes like 4-5 times less time to exec but the queries can be up to 20k chars), so it's up to you.

trinitynet:
@harmdhast why the hell is ur query's 20k 🙃

harmdhast:
@trinitynet 
When I need to update 200 players positions +/- 100 vehicles pos + stats each 30 seconds, I decided to get rid of the multi Update cancer.
So I use a SQL switch. (You have all the data in one query)
Still need to normalize the database, I think I will be able to reduce the queries a little bit.

BoGuu:
@harmdhast 

20k char queries is a serious amount, and it sounds like there would be a better/more elegant way around that. Regardless, this means the only possible issue could be query size, so the other formats aren't affected at all. Instead of the structured text method, you could just concatenate the strings, as they are purely linear `(str _mode) + (str FETCH_CONST(life_sql_id)) + _queryStmt` etc

trinitynet:
@harmdhast dont sound like a good idea to send a massive string to form a query which would be potentially blocking... Yet alone every 30 secs

harmdhast:
@BoGuu 
I won't lie there, I just did a quick fix.
I already started to rework it with string arrays + joinString because it's way faster with huge strings. But as you said I don't think anyone else would need such a fix.
Maybe I will come with something different with the rework of the database.

@trinitynet 
It is.
I had to cut my Update queries in 2 or 3 blocks or more. And the exec time was poor.
Even if the query is 20k chars it can be up to 20 times faster than update statements. 
I did a LOT of benchmarking on this one trust me.

trinitynet:
@harmdhast there are much better ways to handle this, for example have the client send update request after travelling `x` distance. this method hurts my soul

BoGuu:
Of course it'll be quicker, every call is blocking, and if you're updating so often, you could even run into deadlocks :D

harmdhast:
@trinitynet 
Huh 20k char queries is not that huge, some companies use way bigger queries. And it's not hacky at all it's uses a standart sql syntax. The database handle it brillantly because it's meant to handle such queries.
Using an update request on the client adds a loop (or at least more computing) + Useless netword traffic + One call / player. Your method hurt my soul :x

BoGuu:
Indeed, but cutting out the not so brilliant extension system we possess in ARMA :P

I think he is trying to say you could quite easily use a combination. You are updating everything, even if this is needlessly. You could run batch queries as you are now, but only for clients which you have deemed as needing updating. Using a system with distance could be quite minimal on network traffic (go PVEH or something similar), won't have any effect vs. the current situation with RE and whitelists. Then you can guarantee the queries you are sending are valuable, and not junk :)

harmdhast:
@BoGuu 
Using such a system with an update pool and eventHandlers is already planned but not until I kinda sort out the whole database crapy storage thing. It takes time.
I already had distance checks with the old update statements thing that I remove because I changed how the queries are built, that way the query build is faster and the exec time is faster too. I don't care much about network traffic since it's not really between the server and clients.

harmdhast:
Anyway, you can close the PR then.
Not needed. @BoGuu

BoGuu:
You're welcome to change the query constructor to concat instead, no harm done there, quicker regardless (I'd imagine)

harmdhast:
@BoGuu 
You mean like `(str _mode) + ":" (str FETCH_CONST(life_sql_id)) + ":" + _queryStmt`
Imo there's not much difference since format is faster than + if there's 3 strings or more and you had 2 str.

BoGuu:
The times should be almost negligible between them, but this kind of example is the sad truth of SQF. You can do `format ["%1:%2:", ...] + _queryStmt` though, to achieve mostly identical perf. Your PR was to solve issues with char limits, so there is a reason to it :P

I have no problem with this change, and it solves the issue you made the PR to do :)

harmdhast:
Okay.
I may do more PR as I love to spend my day optimizing things so I may found a better or proper way some day. And I hope I could share some database rework with you without much troubles.

BoGuu:
Sounds good :) The *short term* plans for database stuff is removing mres (long over due), and predefineds

harmdhast:
@BoGuu 
Sounds good.
You can leave this PR in pending until we sort things out.

trinitynet:
@harmdhast i went to sleep, Im aware of how much data a SQL DB can push, one of my projects currently have a 12GB DB. That being said the limitation is not SQL but is ArmA and `callExtention` updating a few 20`x` char strings every min or so is going to be better perf in the long run then updating 20k every 30 secs, considering the server also has to collect this data every 30 secs i can imagine this entire setup is not very healthy

harmdhast:
@trinitynet 
It will depend on the number of players and vehicles. And I can have a HUGE amount of both.
The data collect and the query build is optimized as hell and is faster than ever. I can show you the code if you still don't believe me.
Anyway when I save 200 player pos in ~0.08 ticks it's not much of a deal.

Issue #268, opened 2017-02-04
Player restrained can open vehicle inventory


Dim2ke:
### Expected behaviour
When a player is restrained and in a vehicle, he can not open the virtual invetory or gear inventory, if the vehicle is open

### Actual behaviour
When a player is restrained and in a vehicle, he can open the virtual invetory or gear inventory, if the vehicle is open

### Steps to reproduce the behaviour 
1. Connect on COP slot
2. Restrain player
3. Put player in the vehicle and dont lock the vehicle
4. The player restrained, open de vertual inventory with "T" Key and gear inventory with keyobard key.

### My Solution

Add:
```
if (player getVariable "restrained") exitWith {hint localize "STR_NOTF_isrestrained";}; 
if (player getVariable "playerSurrender") exitWith {hint localize "STR_NOTF_surrender";}; // For player surrender and near vehicle.
```
in
```
Framework/Altis_Life.Altis/core/vehicle/fn_vehInventory.sqf
```
Like this:

> http://pastebin.com/JxNg1aPw

---------------------------------------------------
Mission version: 5.0.0

Launcher version: latest

Game version: ~latest~ 1.66

Branch: ~master~ main -- Edit: @Jawshy

mySNAKE:
Good idea! :D How i can contact u for one ask ?

BoGuu:
Thanks for the spot, this would've occurred after disableUserInput was removed from the restrain logic. Adding a check to the keyHandler would be better, if you want to PR this you can, if not, I'll add it

PR #271, opened 2017-02-05
Add check for open inventory (#268)

Dim2ke:
*fn_keyHandler.sqf
*Line: 137

add check if player can open inventory :
Player restrained
Player surrender
life_isknocked
life_istazed


Dim2ke:
That fail, sorry, i have modified line for knock player

PR #300, opened 2017-03-27
Using new UnitLoadout to save gear + Fix #68

### LINKED ISSUE #68: FuelStation does not work with Land_FuelStation_01_pump_F
### Expected behaviour

You should be able to fuel up your vehicle on `Land_FuelStation_01_pump_F` stations.
### Actual behaviour

You can select to "fuel up car", but no dialog opens.
1. Get a car
2. Go to a station that features the `Land_FuelStation_01_pump_F` station (e.g. nearly all in Georgetown, unfortunately).
3. Try to fuel up your car
### RPT and/or extDB2 logs (if applicable)

No logs, but I believe it's because of [this](https://github.com/AsYetUntitled/Framework/blob/master/Altis_Life.Altis/core/shops/fn_FuelStatOpen.sqf#L17).

Update:
I added a `hint` to the `exitWith` above and it indeed showed up, so I believe this is the culprit. I just don't know what the name of the correct feeder would be.

---

Launcher version: 1.5.137506

Game version: 1.62.137494

NiiRoZz:
- Using new UnitLoadout to save gear

Edit: 

- Fix https://github.com/AsYetUntitled/Framework/issues/68 -- @Jawshy

### DIFF
--- a/Altis_Life.Altis/core/functions/fn_loadGear.sqf
+++ b/Altis_Life.Altis/core/functions/fn_loadGear.sqf
@@ -7,6 +7,7 @@
     Loads saved civilian gear, this is limited for a reason and that's balance.
 */
 private ["_itemArray","_handle"];
+_variable = param [0, false, [true]];
 _itemArray = life_gear;
 waitUntil {!(isNull (findDisplay 46))};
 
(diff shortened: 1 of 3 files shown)


Altis_Life.Altis/core/functions/fn_loadGear.sqf
```diff
@@ -7,6 +7,7 @@
     Loads saved civilian gear, this is limited for a reason and that's balance.
 */
 private ["_itemArray","_handle"];
+_variable = param [0, false, [true]];
```
SimZor: Use params here instead

BoGuu:
Still annoys me the command is yet to stop changing, but saves us some hassle. We should do the same to dead gear here

ShadowRanger:
This would break compatibility with the current loadout system. Anyone who updates to this would not be able to keep any saved loadouts at is it uses a completely different format. Is it really necessary?

AmauryD:
I suggest to wait for this ticket to be fixed :
https://feedback.bistudio.com/T127225

Leigham:
this should be good to change i just tested the command over 1000 loops, no visible increase in entities 
[logEntities_20824_17-09-2018_12-08-17.log] Before (https://github.com/AsYetUntitled/Framework/files/2388297/logEntities_20824_17-09-2018_12-08-17.log)
[logEntities_20824_17-09-2018_12-09-45.log] After (https://github.com/AsYetUntitled/Framework/files/2388298/logEntities_20824_17-09-2018_12-09-45.log)

AmauryD:
It was fixed in the last update : https://feedback.bistudio.com/T127225

Leigham:
i just read `should` so I thought I would test it ;)

DomT602:
Unknown repo so cannot resolve conflicts. get/setUnitLoadout will be utilised in 6.x.

PR #302, opened 2017-03-28
UPDATE: move variable type privates instead of array type

SimZor:
Resolves # no issue

- Move away from privates with the array type and do it directly for each variable instead
- This is faster and more optimised

### DIFF
--- a/Altis_Life.Altis/core/actions/fn_buyLicense.sqf
+++ b/Altis_Life.Altis/core/actions/fn_buyLicense.sqf
@@ -6,7 +6,7 @@
     Description:
     Called when purchasing a license. May need to be revised.
 */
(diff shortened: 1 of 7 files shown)


Altis_Life.Altis/core/actions/fn_buyLicense.sqf
```diff
@@ -6,14 +6,13 @@
     Description:
     Called when purchasing a license. May need to be revised.
 */
-private ["_type","_varName","_displayName","_sideFlag","_price"];
-_type = _this select 3;
+_this params ["_type"];
```
BoGuu: ;)?

SimZor:
#310

SimZor:
#311

PR #322, opened 2017-04-20
LoadGear RPT Fix

J4Wx:
- On connect, an undefined variable in expression error will log for all BluFor, this is because loadGear (which contains a call to playerSkins) is called before life_copLevel is defined. This is resolved by moving the call to loadGear to after life_coplevel is defined.
- Also removed a set of unnecessary brackets.


BoGuu:
Thanks! 👍

PR #323, opened 2017-04-22
Old Event-handler we can use this now FiredBIS

TheValthos:

Resolves #<issue ID here>. <If applicable.>

-


ShadowRanger:
Are you sure this event handler is in vanilla Arma 3? I can't find any documentation on it.

TheValthos:
Good point its not documented yet it seems removed until BI push info.

PR #335, opened 2017-05-20
Temporary Debug Menu

imKLM:
- A temporary debug menu to get around BI forcing players to be logged in as admin
- Worth noting this probably has a fair few security issues


SimZor:
Would suggest moving to server side for more security. The exec function that is.

Also, there is no difference between local, global and server? Maybe this is your plan later on though.

SimZor:
Also, don't know why this is temporary, don't think Bohemia has any plans to change this new things implemented to secure the debug. 😆

imKLM:
fair point, server, global and local perform as they should though

ShadowRanger:
I'd suggest just copying the default debug console classes for this as well as the required scripts and editing them. That way you can keep some security compared to this, such as executing code on the server.

imKLM:
cant use the BI function as it runs BIS_fnc_isDebugConsoleAllowed which checks that the source is logged in as admin before executing code

ShadowRanger:
You can use the same functions, just copy the source code of them and make your own, editing them and removing the isDebugConsoleAllowed check.

SimZor:
Don't copy BI, it's not going to help security one bit. My advice, move to server-side, do an admin check, do whatever other checks is needed. Should be secure enough.

SimZor:
And sorry @imKLM, didn't see your remoteExecs

SimZor:
Could use this, https://community.bistudio.com/wiki/remoteExecutedOwner. Checking UIDs on server against a whitelisted array of UIDs. Or getting all UIDs from database on server start with all entries with admin more than 0 and checking that.

Dim2ke:
I have a small suggestion to easily activate and deactivate the menu
Can you add this in Config_Master.hpp
`enable_debugMenu = false; //true for enable debug Menu`
And this at the beginning of fn_adminDebugCon.sqf
`if (LIFE_SETTINGS(getNumber,"enable_debugMenu") isEqualTo 0) exitWith {};`

ShadowRanger:
Just a heads up, this issue has been addressed on the Arma 3 Deveopment branch and will be somewhat resolved in the next Arma 3 update through a 'whitelist' for certain UIDs to have access to the debug console.

https://forums.bistudio.com/forums/topic/140837-development-branch-changelog/?do=findComment&comment=3192280

imKLM:
ah nice, was going to write it similarly myself. guess i'll just wait it out, hopefully the update will be soon

PR #351, opened 2017-07-17
SpyGlass Update 1.72

RosieCode95:
- Update SpyGlass to current arma build


Jason2605:
Arma has now surpassed this version therefore it will be missing patches.

PR #353, opened 2017-07-31
Update fn_setupActions.sqf

SMZAVGSERVERS:
**Changes proposed in this pull request:** 
- adding support for west & independent sides.

### DIFF
--- a/Altis_Life.Altis/core/fn_setupActions.sqf
+++ b/Altis_Life.Altis/core/fn_setupActions.sqf
@@ -6,6 +6,8 @@
     Master addAction file handler for all client-based actions.
 */
 switch (playerSide) do {
+
+    //Actions for civilians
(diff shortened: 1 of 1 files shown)


SimZor:
Why would you add two cases that are not used? 🤔

SMZAVGSERVERS:
@SimZor To make customisation easier ^^
That's not important but it's better than nothing :p

SimZor:
They can just add it in themselves?

SMZAVGSERVERS:
@SimZor Yep, ik but... i'm bored today u know x)

SimZor:
You will also check for unnecessary playerSide, which means you are adding performance decrease in the framework.

BoGuu:
This will become deprecated regardless soon :)

PR #357, opened 2017-07-31
Improving godmod

SMZAVGSERVERS:
Because gods are never tired...

- Desactivate fatigue


SimZor:
What if fatigue is off in config_master, then you are allowing it again...

H0lidays:
^^
If an admin is on-duty, it's easier for them to have the fatigue off.

fresqo:
there should be a check to see if fatigue is on then make the switch.. also in the survival file the godmode var should also bypass food and water consumption.

ColinM9991:
Or, the alternative is to leave this as a no-damage script instead of adding unnecessary complexities.

fresqo:
@ColinM9991 the problem with no-damage is as follows... 

food and water death is setDamage...

<allowDamage>
This command does not prevent object from taking scripted damage such as setHit, setHitIndex, setHitPointDamage or setDamage.

Which means admins will still die from food and water loss even with godmode on.

ColinM9991:
So where does fatigue come in with this?

The correct thing to do here would be to add an event handler for HandleDamage, and remove it when godmode is switched off, but if we're messing with fatigue here in a health script then in months we'll have the same PR for adding unlimited ammo.

Besides, if an admin dies from food and water loss with godmode on, then they're probably using it for too long anyway.

fresqo:
#357 
@ColinM9991 admins never abuse powers what are you talking about?.. what are you talking about unlimited ammo?? godmode does not mean unlimited ammo.. but godmode should mean u DONT die.
i have had issues when i have been on testing things etc gone afk and come back and been dead.

ColinM9991:
`admins never abuse powers`

Conversation over.

This change is a pointless change, I'll leave other editors/admins make the decision.

fresqo:
it was a joke *sigh

SimZor:
@ColinM, tbh I don't think your argument makes sense here. However I do belive if this should be done, you should just terminate the entire thread and start it again when admin goes out of no damage mode, in order for performande optimisation.

ColinM9991:
@SimZor That's why I'm leaving it down to others to decide because I'm biased on changes like this. As I've said, it's a pointless change - from my perspective.

Obviously we are not in the business of telling people how to run their servers, but we are actively maintaining the mission framework to fix bugs and include to features and so it's clear that we do care about the experience which players receive. Why does an admin need infinite fatigue? Infinite health is understandable in a scenario where the admin is addressing several players, but with a teleport function to go to reported players, and the ability to freeze those players, why would an admin need to run endlessly?

sargken:
What is the big deal about this, like lol.

RosieCode95:
closing due to inactivity and op MIA, @SMZAVGSERVERS feel free to reopen if you wish

PR #363, opened 2017-08-11
Update altislife.sql to Create User and Grant Permissions When Run

H0lidays:

Resolves nothing

- Adding a default created user "arma3" for procedures + MySQL queries. Not to be used for administration.


H0lidays:
didn't think 'bout that, thanks

H0lidays:
@ColinM9991 from MySQL doc as I was thinking
"You cannot use wildcards - you have to grant either to all objects (or all objects in a schema) - or then you have to list all objects one by one."

ColinM9991:
That quote is from Stackoverflow about MSSQL, not the MySQL docs.

https://stackoverflow.com/a/14020754/5062791

H0lidays:
@ColinM9991 i think that if it's coming from stack it could be true, isn't it ?

setoy:
Yeah, but we're not using MSSQL here. It might be true for MSSQL, but not for MySQL.

H0lidays:
my bad, anyway got a simplified query that works perfect.

H0lidays:
Any updates on this ? :p

ColinM9991:
I've left reviews which you haven't responded to, you've also merged a branch you didn't need to.

H0lidays:
@ColinM9991 can't see any reviews on this...
Better be updated than outdated, even for one PR

altislife.sql
```diff
@@ -12,12 +12,14 @@ SET time_zone = "+00:00";
...
 --
--- Database: `altislife`
+-- Creates database `altislife` unless it already exists and uses `altislife`
 -- Default Schema
+-- Creates default user `arma3` with password `changeme` unless it already exists
 --
 CREATE DATABASE IF NOT EXISTS `altislife` DEFAULT CHARACTER SET utf8mb4;
 USE `altislife`;
 
+CREATE USER IF NOT EXISTS `arma3`@`localhost` IDENTIFIED BY 'changeme';
```
Jawshy: It would be good for security to append [`PASSWORD EXPIRE`](https://dev.mysql.com/doc/refman/5.7/en/alter-user.html#alter-user-password-management) so the user is forced to change the user password. Unfortunately, [MariaDB](https://mariadb.org/) does not currently (version 10.2.8) support this value. Source: [MDEV-7597 - Expiration of user passwords](https://jira.mariadb.org/browse/MDEV-7597)

altislife.sql
```diff
@@ -210,6 +212,12 @@ CREATE TABLE IF NOT EXISTS `wanted` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
 -- --------------------------------------------------------
+--
+-- Granting permissions to user `arma3`, created above
+--
+
+GRANT SELECT,UPDATE,INSERT,DELETE ON `altislife`.* TO 'arma3'@'localhost';
+GRANT EXECUTE ON altislife.* TO 'arma3'@'localhost';
```
Jawshy: Why is `GRANT EXECUTE` on its own line?

H0lidays: You're right, no need to duplicate this one.

altislife.sql
H0lidays: Could be, but some users will find this kinda useless, and then re-set it up to the original they wrote in the file.

H0lidays:
No problem 👍

PR #375, opened 2017-09-03
Updated: ArmA RPG Life Discord Invite Link

Jawshy:
- Updated expired and invalid links. New invite directs `#general` channel. Does not expire.


jordanlee833:
Maybe this should be done so it is easier for people to find help?

PR #380, opened 2017-09-24
Update fn_gangInvite.sqf

Basti007:
Update fn_gangInvite.sqf
- If player gets invited and don't accept the invite he would not be removed from the gangmember list. <- Fixed this

I can't test it right now. Can please someone test this for me, Thanks!


SimZor:
LGTM. Now correct arguments are passed to the gang update function.

BoGuu:
Thanks! :)

PR #399, opened 2017-11-28
Removal of duped file

DomT602:
Resolves #. 

- Removes a file which is not used, and isn't even present in Functions.hpp


DomT602:
fn_onPutItem also removed as again it wasn't being used, or present in Functions.hpp

Altis_Life.Altis/core/shops/fn_impoundMenu.sqf
```diff
@@ -1,54 +0,0 @@
-/*
-    File: fn_impoundMenu.sqf
```
RosieCode95: you can also remove this from remoteexec.txt IIRC its in there

DomT602: This file was duped, as it can be found in dialog/functions where the actual one is used, so its still relevant to be in remoteexec.txt :)

RosieCode95: an no probs, my bad was very much a zombie

PR #402, opened 2017-12-15
Put/Take EventHandler - #361

### LINKED ISSUE #361: Weapon dupe
### Expected behaviour
If you put weapon inside vehicle and press alt+f4 then re log you should not have your weapon back in your hand

### Actual behaviour
if you put your weapon inside vehicle and press alt+f4 while inventory open, then rel log your weapon duplicated, in your hand and inside vehicle inventory

1. open vehicle inventory
2. put your weapon inside vehicle inventory and press alt+f4 while open inventory
3. re log then your weapon in your hand and in vehicle inventory

----

Mission version: 5.0

Launcher version: 1.5.142122

Game version: 1.74.142559

Haizanjs:
If someone could please test.

Resolves #361 

- Force Sync upon taking/putting in a Physical Item from a weapon holder container e.g. vehicle inventory.


jackmsimo:
@SimZor This should be what you were looking for [here](https://github.com/AsYetUntitled/Framework/pull/402/files#diff-20e1a4a9d1f232a52e15978261160597)

DomT602:
If your just wanting to update gear, there's no point doing a full updateRequest for it, you can use [3] call life_fnc_updatePartial; instead. Also this'll need to be tested as this could put a lot of stress on the network as putting/taking items is very common and it may not be worth it performance wise

SimZor:
@DomT602 this is true, however, it will not be a problem if we just implement an eventhandler handling all the update requests in a queue.

BoGuu:
What @DomT602 said is important (the full sync is not needed). However, this situation really needs to be taken into consideration carefully. For the past few years AL has moved over to a 'sync when anything happens' mentality, and as a result, the network traffic is through the roof. The biggest culprit for this is updates (you're welcome to verify this).

I think this should be on hold until such a solution is found, or not.

Thanks for the PR :+1:

jackmsimo:
@BoGuu Other instances of systems using a full sync when need to save data but not needed a full sync?

BoGuu:
@jackmsimo 

Here is the list of full updates:

https://github.com/AsYetUntitled/Framework/search?utf8=%E2%9C%93&q=sock_fnc_updaterequest&type=

So anything where a certain variable is not affected in the calling function. Such as seize client, etc.

jackmsimo:
Welp even for fn_revived, only the alive column / bank column needs updating....

Jason2605:
Issue #361 should now be resolved.

Thanks for the PR regardless!

PR #403, opened 2017-12-19
Update fn_calWeightDiff.sqf

AmauryD:
Function is faster and cleaner.
I've got performance issues with the old function when the _num var was too big (function was taking ages to complete).

### DIFF
--- a/Altis_Life.Altis/core/functions/fn_calWeightDiff.sqf
+++ b/Altis_Life.Altis/core/functions/fn_calWeightDiff.sqf
@@ -7,23 +7,15 @@
     Calculates weight differences in the _cWeight (current Weight) against the _mWeight (Max Weight)
     Multiple purpose system for this life mission.
 */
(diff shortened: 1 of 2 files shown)


DomT602:
@AmauryD could you swap the spaces to tabs for travis (can be done thru your text editor), also changing this to just: call life_fnc_virt_buy (https://github.com/AsYetUntitled/Framework/blob/master/Altis_Life.Altis/dialog/shop_items.hpp#L110) should hopefully ensure this bug is quashed: https://github.com/AsYetUntitled/Framework/issues/405

AmauryD:
Thank's. I did it , took me 1 hour to find the indent options in my text editor.

DomT602:
Thanks for doing that, at least you'll know where it is in the future :)
@Jawshy this is ready for merging, been tested by me, Bailey & MrWeyland, and it successfully dealt with https://github.com/AsYetUntitled/Framework/issues/405

PR #406, opened 2018-01-01
Fix possible bug with fn_calWeightDiff

DomT602:
Resolves #. 

- If an item weights 0 (not in default configs), zero divisor error occurs, this fixes that.


AmauryD:
Maybe return -1 will be better ?

DomT602:
If it returns -1, it will mean that no items can be bought e.g
_diff = [_type,parseNumber(_amount),life_carryWeight,life_maxWeight] call life_fnc_calWeightDiff;
if (_diff <= 0) exitWith {hint localize "STR_NOTF_NoSpace"};
As -1 is lower then 0, it will exit, even though with a weight of 0 an infinite amount of the item can be bought, so basically we're just exiting the script with the value they put in, as there is no need for any maths.

PR #420, opened 2018-02-05
Update fn_broadcast.sqf

Kapupc:
localize message with none argument is not working due to _arr is not passed by remoteExecutedOwner

Resolves #. 

- 


DomT602:
lgtm, if you could just remove the parentheses around the 2 'switch (_type) do {''s and then use the private keyword before _msg on line 17 and then this file will be brought up to better standards.

Jason2605:
Thanks

PR #435, opened 2018-04-16
Oh! A Mistake!

Ermen07:
Fix a mistake!

### DIFF
--- a/Altis_Life.Altis/core/functions/fn_numberText.sqf
+++ b/Altis_Life.Altis/core/functions/fn_numberText.sqf
@@ -11,7 +11,7 @@
     Returns:
     STRING
 */
-private ["_number","_mod","_digots","_digitsCount","_modBase","_numberText"];
+private ["_number","_mod","_digits","_digitsCount","_modBase","_numberText"];
 
 _number = [_this,0,0,[0]] call bis_fnc_param;
 _mod = [_this,1,3,[0]] call bis_fnc_param;


AWildTeddyBear:
This wasn't a breaking issue actually as 
```sqf
_digits = _number call bis_fnc_numberDigits;```

Defined _digits on. But this whole file needs a re-write tbh, I'll work on this soon

AWildTeddyBear:
I've done a re-work of it, it should have the exact same effect and output as the original script, just updated to newer sqf standards.
https://github.com/AsYetUntitled/Framework/commit/5d56a08499572e44b9b628fdb8b445295b30a188

BoGuu:
@Ermen07 @AWildTeddyBear 

Before we pull any of these changes, it might be worth noting the following:

fn_numberDigits was a carbon copy of BIS' Karel Moricky's numberDigits (present in the game). I just double checked, and it has since changed since it was copied into AL. Perhaps it would be worth checking if his new version is better than this, and moving all usage to the BIS_* function (which it should've been originally)?

It seems the new vers. doesn't use `str`

Leigham:
Seems good to me, just tested, using the arma3 user with extdb, no errors.

DomT602:
As the originating repository no longer exists, this PR will be closed as no further edits can be made. While the mistake is present, the whole file could be optimised in a PR fixing this.

PR #457, opened 2018-06-03
Corrected misleading reference to lines

Logfro:
Pretty sure it is self explanatory


Jason2605:
Thanks for the PR!

fresqo:
dem sneaky misleads

PR #462, opened 2018-06-25
Removed public slots and edited the admin access

gitAlexDE:
Public slots and admin access (fn_initCop.sqf)
- public slots removed
- admin query edited
 -> If you want allow your admins to use the slots then you need to edit "admin_slotAccessAllowed_COP" in the Config_Master.hpp to true.

admin access (fn_initMedic.sqf)
- admin query edited
 -> If you want allow your admins to use the slots then you need to edit "admin_slotAccessAllowed_MEDIC" in the Config_Master.hpp to true.

admin access (Config_Master.hpp)
added
- admin_slotAccessAllowed_COP
- admin_slotAccessAllowed_MEDIC

- [yes] I have tested my changes and corrected any errors found

### DIFF
--- a/Altis_Life.Altis/config/Config_Master.hpp
+++ b/Altis_Life.Altis/config/Config_Master.hpp
@@ -162,6 +162,10 @@ class Life_Settings {
     disableCommanderView = true; //false - Group leaders can access the commander view. true [default] - Group leaders cannot access the commander view.
                                  //Commander/tactical view is accessed via pressing . [NUM] by default. It raises the camera significantly higher and steeper above the player in order to give a boarder tactical view of the surrounding area.
 
+    /* Admins access to medic/cop slots */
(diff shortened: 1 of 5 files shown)


AWildTeddyBear:
I think this would be moreof a personal change and not something for the official repo. ;P

gitAlexDE:
It might be a personal but with this change you can decide what you want, so i think it isn't a bad idea for the official repo.

AWildTeddyBear:
The repo is designed for "you can decide what you want" tho. 🤔

ToxicRageTv:
why not add an option in the master_config for public cops

AWildTeddyBear:
Because it's a framework designed to be changed to user's needs? 🤔

gitAlexDE:
I understand your point of view, but then some decisions in Config_Master.hpp would not even exist. We should probably keep up with the times and adjust the framework accordingly. I doubt that the versions are only there to publish bug fixes.

PR #414, opened 2018-01-06
SQL-Custom

Jason2605:
# SQL-Custom
#### Changes proposed in this pull request: 
- Change current SQL queries to prepared statements
- Remove all mres functions

- [x] I have tested my changes and corrected any errors found (life_server, haven't been able to test HC as of this moment)

PR #468, opened 2018-08-17
optimize Mysql fnc_numberSafe

dedmen:
input `[1234241.1231231231]`

Old: 0.361ms output: `"1234241"`
New: 0.0057ms output: `"1234241"`

I don't know what _mod parameter is supposed to do though. No values that I tried for it changed anything.

Resolves #. 

- 

### DIFF
--- a/life_server/Functions/MySQL/fn_numberSafe.sqf
+++ b/life_server/Functions/MySQL/fn_numberSafe.sqf
@@ -11,18 +11,6 @@
     Returns:
     STRING
 */
-private ["_number","_mod","_digots","_digitsCount","_modBase","_numberText"];
+params ["_number"];
 
-_number = [_this,0,0,[0]] call bis_fnc_param;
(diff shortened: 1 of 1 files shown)


ConnorAU:
Yea don't worry about _mod, it never needed to be in this script. It's purpose is to add a character every _mod digits to make the number easier to read (eg: 1234657 -> 1,234,567) but it added no character [here](https://github.com/AsYetUntitled/Framework/pull/468/commits/c3a657ab74728e2d39b7ecc4b2ec6741deadcd65#diff-4f15027822f236ae5a32a9701939788dL26). This script was just a copy paste of BIS_fnc_numberText so that's where _mod came from.

DomT602:
Like with #414 this function will be removed but this is fine for now.
Thanks for the PR :)

Jason2605:
#414 Has been merged and the mres functions have been removed, thanks for the PR however!

PR #475, opened 2018-08-31
new 3d vehicle store

### LINKED ISSUE #154: Undefined variable life_preview_3D_vehicle_object
Since I cannot commit changes I would like to inform a fix for this error.

You need to add `waitUntil {!isNil "life_preview_3D_vehicle_object"};`
Before the line `waitUntil {!isNull life_preview_3D_vehicle_object};`
In the file fn_vehicleShopInit3DPreview.sqf.

imKLM:
Resolves #154 as it's an entirely new store

Adds a new 3d object preview to replace the current 3d vehicle store preview.
This one may also work with other object not just vehicles (untested with weapons).
this one adds new functionality such as click&drag to rotate the object and also mouse wheel zooming.
Have tested the zooming with all vehicles which are in the default civilian stores and the camera does not clip into any of the vehicles, can adjust offset if needed.


Altis_Life.Altis/core/shops/fn_vehicleShopMenu.sqf
```diff
@@ -64,8 +64,4 @@ ctrlShow [2304,false];
     };
 } forEach _vehicleList;
 
-if (LIFE_SETTINGS(getNumber,"vehicleShop_3D") isEqualTo 1) then {
-  [] call life_fnc_vehicleShopInit3DPreview;
```
imKLM: the init is now ran in the dialogs onLoad attribute so this is no longer needed

PR #487, opened 2018-09-22
Update fn_vehicleShopBuy.sqf

Logfro:
- Fixes a bug with which you can not store a vehicle when it is bought.


BoGuu:
Thanks for the PR 👍

PR #489, opened 2018-09-30
Add contracting section to briefing.sqf

Turdant:
Added a contracting setting in police rules

Resolves #. 

Added a contracting section to the briefing, which was referred to on line 246, but didn't actuallyexist.


DomT602:
Hi @Turdant. Firstly thanks for the PR! Secondly thanks for spotting the inconsistency as it has managed to be missed. After discussion we have decided to remove the reference to contracting instead of this, as we would rather server owners add this as a feature, rather then it be in as default as this is just a Framework after all. Thanks again!

PR #490, opened 2018-10-01
Blasting charge improvements

robinwithes:
Changes proposed in this pull request: 
If local player is in a valid altis life gang group then the bank timer will also show up on every online group member of that player. I also added a sidecheck in the script since i did not see any in both the blastingcharge and useitem script.

- I have tested my changes and corrected any errors found


DomT602:
Due to the fact that the originating repository is deleted, and the changes were not made (and I can't change this PR) this will be closed. Feel free to re-make the PR with the requested changes.

PR #491, opened 2018-10-01
Knockout improvements

robinwithes:
Instead of immediatly being thrown to the floor in a very static animation, you now first ragdoll onto the floor and transition into the animation. Only 3 lines of code and wont break anything. Had this on our server for 5 months without any issues.


DomT602:
Thanks for the PR, however this would be up to individual servers to further customise the default  knockout feature.

Issue #111, opened 2016-09-25
Skill/Proficiency System


Repentz:
Have a system implemented that gives the basics to rewards for repeating a process

- Fishing
- Mining
- Woodcutting
- Impounding
- Ticketing

Give perks for these per level, such as faster processing, use of better tools per level like drills for mining rather than pickaxes, chainsaws, etc.

Give more % of ticket to actual cop per ticket, jail, etc.

Just a suggestion, I've requested it be added to the framework before because I think it would add alot to the RP aspect of AL. You can also have skill system = True or False for people who don't want it.

policeman18:
+1

Latouth:
+1

fresqo:
IMO, This should either be IN as a new standard feature or not put in. Please don't put as option as its too much code to just sit around if not used.

tkcjesse:
Yeah. Exactly. I personally think it shouldn't be in. I know multiple people who hate it and also people who like it.

MJRamon:
Players like progression. Players like to be awarded for repetative work. This is good and healthy approach and has to be built-in.

BoGuu:
Despite the fact there is two pull requests for arguably 'non Framework' things, I do not stray away from the fact that this is still a Framework, and features such as this do not entirely belong on it.

There are lots of servers out there that have put countless hours of work into their code base, making things such as this. So for it to become a default in the Framework (for every server to have), would be disappointing. There is a reason why some servers are good, and some are bad, a lot of that boils down to the content, I don't feel it's right to let somebody take 5 minutes to achieve what took some months.

tkcjesse:
Hah. You know i agree dawg.

fresqo:
most mmorpg/rpg games have a leveling system of some sort (even just a general playing lvl) i think arma life mods kinda need it tbh.

Emma-L:
[...]

PR #493, opened 2018-10-01
Added panic button v-item for police

robinwithes:
Added a v-item called panic button which when used sends a emergency broadcast to the police as well as marking the location of the cop in distress to them.


robinwithes:
Not sure what the checkstyle issue is here

imthatguyhere:
> Not sure what the checkstyle issue is here

Put a space between `if` and `(` on the two lines noted.

SimZor:
Yeez

DomT602:
Hi @robinwithes, firstly thanks for the PR, when it comes to this feature; we believe it be left to server owners to implement this how they want, in order to make their server unique. This would be more suited to being on a separate public repository where you could release these scripts as small addons that server owners can add if they want. We have had similar situations, namely this https://github.com/AsYetUntitled/Framework/issues/111#issuecomment-252308203.
With this in mind, I'll be closing this PR.

PR #496, opened 2018-10-19
Kron & init cleanup 

DomT602:
Resolves N/A. 

- Replaces KRON functionality with newer commands
- Cleans up init related


moeck:
if I try to acess the trunk it ist not possible to get things out. Error message "wrong number format"

DomT602:
Good find, I forgot to remove the str surrounding _x for the isNumber function, that’ll be the issue

moeck:
you are right, I changed it in the life_server\functions.sqf and it works for me

Jason2605:
Good spot thanks, I've updated the main branch now!

PR #499, opened 2018-11-03
Add Jail Timer to config master for easier access

IAmProphet:
Basically I just changed it to grab from a variable rather than a fixed number and the variable is set in the config_master so for people new to the framework they can modify it in the config_master.


Altis_Life.Altis/core/civilian/fn_jailMe.sqf
```diff
@@ -14,7 +14,8 @@ params [
 ];
 
 
-if (_bad) then { _time = time + 1100; } else { _time = time + (15 * 60); };
+_jailmultiplier = LIFE_SETTINGS(getNumber,"jail_time_multiplier");
```
DomT602: Pop private before this please!

Altis_Life.Altis/core/civilian/fn_jailMe.sqf
```diff
@@ -14,7 +14,8 @@ params [
...
+if (_bad) then { _time = time + (_jailmultiplier * 60) + 900; } else { _time = time + (_jailmultiplier * 60); };
```
DomT602: Could this if be moved to a multi-line if, for standardisation sake

DomT602:
Thanks for the PR @IAmProphet

PR #509, opened 2018-12-20
Update briefing.sqf to remove contracting mention

imKLM:
- Removes the mention of contracting as it should be a feature servers decide if they want or not.


DomT602:
Cheers @imKLM!

PR #510, opened 2018-12-22
Fixed: Instant Revival if outside distance

Scarso327:
We have a drag option on our server and when someone was taken outside this range the isNil check didn't work and so it would instantly revive the player without stopping it at the line 67 exitWith.


Scarso327:
Dunno if I did a good or not :p Just update stuff from what I interrupted your review to mean.

DomT602:
@Scarso327 thats pretty good so far 👍 Just one last thing, we can change; private _target = param [0,objNull,[objNull]];
to; params [
    ["_target",objNull,[objNull]
];

Scarso327:
Done. Is parmas faster or something?

Leigham:
@Scarso327  Params is newer.

Leigham:
`if !(params [
    ["_target",objNull,[objNull]]
]) exitWith {};`

you had an error, if you use this, you can also remove the `if (isNull _target) exitWith {}` line

Leigham:
@Scarso327  please read what I put.
you cant just remove the line

Scarso327:
Sorry, I am jumping between server work and this. What you said was hard to understand until the "also" part

Leigham:
params returns a boolean, meaning, if the data doesn't match the type, or its default, it will return `true` so if you put `if !(params [blah]) exitWith` it will exit the script if it is default values.

Scarso327:
But is isNull check before already did that?

Leigham:
why have an error uneeded line ?

Scarso327:
OH, I UNDERSTAND

Jason2605:
Good spot 👍

PR #519, opened 2018-12-26
Spike Strip Optimization

### LINKED ISSUE #212: Placing spikestrips - can only place one
after placing a spikestrip i want to place another one
it tells me i have already a spikestrip in deployment

you cant even place another one when a vehicle drives over the first one and it gets destroyed

you need to relog then you can place another one

happens in 4.4r4 and the spikestrip script is similar to dev

its no real bug its the way the script is done...
since it checks in line 10 if there is a pickup addaction you cant place a second one
and this addaction is not removed when the vehicle drives over it and destroys the spikestrip

Propheto:
Resolves #212 

Optimizes all the spike strip files, also makes them universal so you can place as many as you want and pick up as many as you want.


DomT602:
Happy enough with code, just need to sort Travis out.

imKLM:
Perhaps have a config based limit? with something like -1 being for unlimited usage.
It's a good idea though! I can just see it being abused easily

DomT602:
I know this was from a while ago, but thoughts on klm's changes? I think its a logical step @IAmProphet.

Altis_Life.Altis/core/items/fn_spikeStrip.sqf
```diff
@@ -6,28 +6,22 @@
...
-_spikeStrip = "Land_Razorwire_F" createVehicle [0,0,0];
+private _spikeStrip = "Land_Razorwire_F" createVehicle [0,0,0];
 _spikeStrip attachTo[player,[0,5.5,0]];
 _spikeStrip setDir 90;
-_spikeStrip setVariable ["item","spikeDeployed",true];
 
-life_action_spikeStripDeploy = player addAction[localize "STR_ISTR_Spike_Place",{if (!isNull life_spikestrip) then {detach life_spikeStrip; life_spikeStrip = objNull;}; player removeAction life_action_spikeStripDeploy; life_action_spikeStripDeploy = nil;},"",999,false,false,"",'!isNull life_spikestrip'];
+life_action_spikeStripDeploy = player addAction[localize "STR_ISTR_Spike_Place",{detach life_spikeStrip; life_spikeStrip = objNull; player removeAction life_action_spikeStripDeploy; life_action_spikeStripDeploy = nil;},"",999,false,false,"",'!isNull life_spikestrip'];
 life_spikestrip = _spikeStrip;
-waitUntil {isNull life_spikeStrip};
```
DomT602: This waitUntil is needed to halt the script for the current functionality, aswell as removing the action. Another method would be to make a separate function for when the spikes are placed, running the remaining code then.

Turdant:
`16:01:57 Error in expression <[getPos player,["Land_Razorwire_F"],5]) params ["_spikes",objNull,[objNull]]; !i>
16:01:57   Error position: <params ["_spikes",objNull,[objNull]]; !i>
16:01:57   Error Type Object, expected Array,String`
Getting this error with these changes

Leigham:
> params ["_spikes",objNull,[objNull]];

the issue is here 
should be 
`params [
    ["_spikes", objNull, [objNull]]
];`

DomT602:
Unknown repo so this PR will be closed. Spikes will be reviewed for 6.x.

PR #526, opened 2019-01-18
Add custom distance medics nears at DeathScreen

Joaquinee:
Resolves #. 

- 

### DIFF
--- a/Altis_Life.Altis/core/medical/fn_deathScreen.sqf
+++ b/Altis_Life.Altis/core/medical/fn_deathScreen.sqf
@@ -1,3 +1,4 @@
+#include "..\..\script_macros.hpp"
 /*
     File: fn_deathScreen.sqf
     Author: Bryan "Tonic" Boardwine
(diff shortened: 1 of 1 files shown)


Jason2605:
Hi @Joaquinee. I'd first like to thank you for the contribution, however there are a few things to address.
When you want to create a pull request, you do so with the aim of changing one thing (ideally), be that creating a new feature, or a bug fix, however this needs to be completed in its entirely for a single pull request. What i mean by this is i've seen you open a request [here](https://github.com/AsYetUntitled/Framework/pull/527) which is required functionality for this change to work. Instead what you would do, is fork the main repository, create a branch, commit all of your changes to this branch and finally test those changes locally. Once you are happy your changes are working as expected, you can create a pull request here, filling out the template to explain a little bit about what you've done and why.

Tl;dr: Im going to close the second PR and, if you could, can you commit the change to this branch instead. Thanks.

DomT602:
As this PR alone will not work, and due to inactivity, this will be closed. Feel free to make the PR (following normal protocol) in the future, if you need any advice on doing this feel free to join us on discord.

PR #529, opened 2019-01-22
Save All Guns

### LINKED ISSUE #520: 2nd Main Weapon (Rocket-Launcher) are not saving on the Database
Hello,

The 2nd Main Weapons (Rocket-Launcher) are not saving on the Database. 
I have already tried adding it in the fn_loadGear.sqf and fn_saveGear.sqf, but then I get an error from the fn_loadGear.sqf that the gear can´t load it from Database and I´m not coming into the Spawnscreen. Have any an Fix of these please? 

Thanks in advance, I need it :D

Propheto:
Resolves #520 
- Saves the secondary weapon (RPG) to the database

### DIFF
--- a/Altis_Life.Altis/core/functions/fn_loadGear.sqf
+++ b/Altis_Life.Altis/core/functions/fn_loadGear.sqf
@@ -7,9 +7,10 @@
     Loads saved civilian gear, this is limited for a reason and that's balance.
 */
 private _itemArray = life_gear;
+[_itemArray select[0,10],_itemArray select 10] params [["_gear",[],[[]]],["_yItems",[],[[]]],"_handle"];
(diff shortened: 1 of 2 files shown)


ConnerMerlin:
The way we have it working is we check the format coming from the database and if its the old format we use that once then use the new method to store it back to the database. This stops the problem where people will lose gear after logging in after changing this.

BlackFisch:
You could do it in a bit easier way if you don't append the y-items to the same array as the gear. I'm leaving my files here for reference.

My files: https://gist.github.com/BlackFisch/9bd18a7d7d9849d77c3feec4ca4c1666

DomT602:
Unknown repo so cannot edit, get/setUnitLoadout will be implemented for 6.x.

PR #536, opened 2019-02-07
SQL Upgrade

imthatguyhere:
Resolves None. 

- Changes some VARCHAR sizes for consistency.
- Adds foreign keys.
- Updates Primary Keys.
- Adds new indexes.

This should be a drop in replacement with no issues, and is working fine in my server currently.
The changes made allow the database to be handled easier with relations, and makes things a bit more consistent.


Jason2605:
Firstly thanks for the PR, if you could add the changes we discussed via PMs that would be great!

imthatguyhere:
Tested and all good.

PR #537, opened 2019-02-09
Emergency lights update and default

Turdant:
Resolves #. 

Cop changes:
Adds a default position for cop cars
Fixes some wording

Medic changes:
Make it work the same as fn_copLights.sqf
Adds a default
Fixes some wording


Altis_Life.Altis/core/cop/fn_copLights.sqf
```diff
@@ -41,6 +41,14 @@ private _offset = switch (typeOf _vehicle) do {
     case "B_Heli_Transport_01_F": {
         [-0.5, 0.0, 0.81];
     };
+    default: {
```
commy2: Wrong syntax. It is `default {` without `:`.

commy2: Yes. And those are my thoughts. I can't comment in that channel :(

Turdant: Fixed!

PR #539, opened 2019-03-04
new option added widthRailWay to dialog

moeck:
with the update to 1.90 a new option to Life_RscMapControl is added

- added the new dialog option

- [x ] I have tested my changes and corrected any errors found


C0kkie:
Just to add: As far as i know is 4 default value in arma

moeck:
actually 1 is working for me ;-)

C0kkie:
That is possible. I just said that 4 is default

BlackFisch:
Can confirm, default value is 4 - shoudle be changed to that to keep stuff unifom
See: https://gist.github.com/BlackFisch/86973392c6bdb6dba9677d29200b37dd#file-ui_f-config-bin-rscmapcontrol-L51
(pulled it out of config like 2min ago, so should be the latest information)

moeck:
changed it to 4

Altis_Life.Altis/dialog/spawnSelection.hpp
```diff
@@ -47,6 +47,7 @@ class life_spawn_selection {
             maxSatelliteAlpha = 0.75;//0.75;
             alphaFadeStartScale = 1.15;//0.15;
             alphaFadeEndScale = 1.29;//0.29;
+            widthRailWay = 4;
```
DomT602: As we're inheriting from Life_RscMapControl I don't see the need to repeat this, otherwise thanks for the PR.

DomT602:
@moeck Thanks for changing it, but you've managed to delete the whole file D:

moeck:
shit happens, now it should be ok

PR #542, opened 2019-03-25
Rework fn_fetchVehInfo.sqf

### LINKED ISSUE #541: Vehicle Shop: Number of passengers not displayed correctly
### Expected behaviour
Vehicle Shop info should display number of seats correctly

### Actual behaviour
As soon as a vehicle has Turrets defined, i.e. for Firing from Vehicles, the number is not correct

Open the Vehicle Shop
_or_
call life_fnc_fetchVehInfo with a classname of a vehicle

----

Mission version: every version up to date

Launcher version: 1.5.145347

Game version: 1.90.145471

BlackFisch:
Resolves #541

- update function to return correct passenger number (see #541)
- removed unncessary "predefinitions" (engine default values should do the job...)
- moved away from overusing `FETCH_Config2`-macro, instead preload the config and fetch values (slight performance increase, could affect delays and such -> https://imgur.com/a/M0p9jRP)
- rewrite function to use `params` instead

### DIFF
--- a/Altis_Life.Altis/core/functions/fn_fetchVehInfo.sqf
+++ b/Altis_Life.Altis/core/functions/fn_fetchVehInfo.sqf
@@ -1,7 +1,7 @@
-#include "..\..\script_macros.hpp"
 /*
     File: fn_fetchVehInfo.sqf
     Author: Bryan "Tonic" Boardwine
+    Edited: blackfisch
 
     Description:
     Used in returning information about a vehicle from Config >> "CfgVehicles"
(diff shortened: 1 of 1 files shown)


Altis_Life.Altis/core/functions/fn_fetchVehInfo.sqf
```diff
@@ -21,37 +21,28 @@
     11: Max Horse power
     12: Fuel Capacity
 */
-private ["_class","_scope","_picture","_displayName","_vehicleClass","_side","_faction","_superClass","_speed","_armor","_seats","_hp","_fuel"];
-_class = [_this,0,"",[""]] call BIS_fnc_param;
+if !(params [
+    ["_class","",[""]]
+]) exitWith {[]}; //Bad params
```
commy2: Should be:
```sqf
params [["_class","",[""]]];
```
Next line already exits for bad `params`.

Also, the params return value generally is not very useful, as it reports `false` for at least one or more undefined (optional) arguments. Should be always ignored imo.

Altis_Life.Altis/core/functions/fn_fetchVehInfo.sqf
```diff
@@ -21,37 +21,28 @@
...
+private _scope = getNumber(_config >> "scope");
+private _picture = getText(_config >> "picture");
+private _displayName = getText(_config >> "picture");
+private _vehicleClass = getText(_config >> "vehicleClass");
+private _side = getNumber(_config >> "side");
+private _faction = getText(_config >> "faction");
+private _superClass = inheritsFrom _config;
+private _speed = getNumber(_config >> "maxSpeed");
+private _armor = getNumber(_config >> "armor");
+private _seats = getNumber(_config >> "transportSoldier") + count ("true" configClasses (_config >> "Turrets")); //number of seats = number of passengers + number of vehicle turrets
```
commy2: Before it didn't count the driver seat, gunner seat, commander seat, and any other turret seat.
Now it counts every seat, including drone AI, minus one for the driver.

Both seem strange, so I am not sure what this is supposed to be.
The best way to count all seats usually is to use `count fullCrew [_vehicle, "", true]`, but that doesn't work on classnames.

BlackFisch: I get that one with AI, but the Problem with not counting turrets is, that with most vehicles the number of seats is otherwise totally incorrect. 
Example: MB4 WD - old function returns 0, actual number of seats is 3 + 1 driver. new function returns 3
Prowler LMG - old function returns 0, actual number of seats is 4 + 1 driver, new functions returns 4

Most of the vanilla vehicles now have "Firing from vehicle" implemented, which means they changed "normal" seats to turrets to accomplish this. That's why I changed it to count turrets too.

Altis_Life.Altis/core/functions/fn_fetchVehInfo.sqf
BlackFisch: removed the exit

Altis_Life.Altis/core/functions/fn_fetchVehInfo.sqf
```diff
@@ -21,37 +21,28 @@
     11: Max Horse power
     12: Fuel Capacity
 */
-private ["_class","_scope","_picture","_displayName","_vehicleClass","_side","_faction","_superClass","_speed","_armor","_seats","_hp","_fuel"];
-_class = [_this,0,"",[""]] call BIS_fnc_param;
+params [
+    ["_class","",[""]]
+];
 if (_class isEqualTo "") exitWith {[]}; //Bad class passed.
 if (!isClass (configFile >> "CfgVehicles" >> _class)) exitWith {[]}; //Class doesn't exist in CfgVehicles
```
DomT602: My only comment is regarding this and line 31, just for the sakes of not repeating code, you could define _config beforehand, then test it with isClass (this would make it slightly slower if there is no class, but in most cases it'll be a tiny gain.).

PR #544, opened 2019-04-03
Fixes a security issue

C0kkie:
- In general it is theoretically possible to execute every script on the server via a hackmenu or some other shit, because bis_fnc_execvm was approved for some reasons in the cfgremotexec, which is not needed.

- [ ] I have tested my changes and it worked as expected, the only thing what i couldn't test is with headless client


Altis_Life.Altis/initPlayerServer.sqf
```diff
@@ -1,10 +0,0 @@
-#include "script_macros.hpp"
-/*
-    File: initPlayerServer.sqf
-    Author:
-
-    Description:
-    Executed only on server when a player joins mission.
-*/
-if (LIFE_SETTINGS(getNumber,"player_deathLog") isEqualTo 0) exitWith {};
```
DomT602: We've lost whether to enable/disable it with your new fix.

C0kkie: Done

DomT602:
@C0kkie sorry for the lack of activity, as pointed out in the discord 'BIS_fnc_execVM' also needs to be removed from SpyGlass. Thanks for the previous fix though!

C0kkie:
Should be done @DardoTheMaster

Turdant:
You forgot `bis_fnc_execvm_meta` on line 163 in _Config_SpyGlass.hpp_

C0kkie:
done

PR #559, opened 2019-05-10
Chinese set as Original??

tomasalves8:
Resolves #. 

- Chinese was set as the Original in the string table file, so I changed it.


Jason2605:
Thanks for the PR!

PR #571, opened 2019-07-22
Fixing house door lock/unlock animation

procyonro:
This should fix the house door lock/unlock animation, as it was broken before.

- [ x ] I have tested my changes and corrected any errors found


DomT602:
When tested the doors didn't work, on another note with the help of @imthatguyhere,_veh animateSource [format ["Door_%1_sound_source", _door], 0]; works

Leigham:
bis_fnc_door should work.

AngryHerring:
> When tested the doors didn't work, on another note with the help of @imthatguyhere,_veh animateSource [format ["Door_%1_sound_source", _door], 0]; works

_veh animateSource [format ["Door_%1_sound_source", _door], 0];

imthatguyhere:
> bis_fnc_door should work.
After you mentioned it, we looked at that function and it would be more efficient not to use it. The only advantage would be it turning the doorknob too, but we could just as easily add that directly without the use of param and the extra function.

Leigham:
> > bis_fnc_door should work.
> > After you mentioned it, we looked at that function and it would be more efficient not to use it. The only advantage would be it turning the doorknob too, but we could just as easily add that directly without the use of param and the extra function.

was just an idea, as its a function already in the game, plus is should work with all `correctly configured` buildings.

PR #572, opened 2019-07-25
Add unsafeCVL parameter to description

BlackFisch:
- adding ```unsafeCVL = 1``` to description.ext to allow createVehicleLocal to be executed on multiplayer servers

This is necessary because of the Arma 1.95 Update released a few hours ago.
For reference see:
- https://community.bistudio.com/wiki/Description.ext#unsafeCVL
- https://community.bistudio.com/wiki/createVehicleLocal


BoGuu:
Is this actually required?

>This command can be restricted in multiplayer by setting unsafeCVL param to 0. Default is unsafeCVL = 1;. unsafeCVL  Introduced with Arma 3 version 1.95

On the wiki page you linked

BlackFisch:
According to this: https://community.bistudio.com/wiki/Description.ext#unsafeCVL default is 0 which means disabled in multiplayer. Was also confirmed in the german forums https://native-network.net/ where I learned about the "fix"

PR #576, opened 2019-09-11
server function cleanup (Changing all BIS_fnc_Param with Param/Params)

koalas11:
Resolves #Function with BIS_fnc_Param. 

-  Changing all BIS_fnc_Param with Param/Params and removing or adding missing variable or useless one + optimization. Checked for errors when starting the server no errors found but still need a better check

- [X] I have tested my changes and corrected any errors found


life_server/Functions/Gangs/fn_queryPlayerGang.sqf
```diff
@@ -5,7 +5,7 @@
     Description:
     Queries to see if the player belongs to any gang.
 */
-private ["_query","_queryResult"];
+private ["_query","_queryResult","_tmp"];
```
DomT602: While we're here, it might make most sense to remove all private ARRAY usage

life_server/Functions/Gangs/fn_removeGang.sqf
```diff
@@ -6,7 +6,7 @@
     Description:
     Removes gang from database
 */
-private ["_group","_groupID"];
+private ["_group","_groupID","_result"];
 _group = param [0,grpNull,[grpNull]];
```
DomT602: this can become params [
["_group",grpNull,[grpNull]]
];

life_server/Functions/MySQL/fn_insertVehicle.sqf
```diff
@@ -5,7 +5,7 @@
     Description:
     Inserts the vehicle into the database
 */
-private ["_query","_sql"];
+
 params [
```
DomT602: Could these be given defaults

life_server/Functions/Gangs/fn_removeGang.sqf
koalas11: Are you sure that even for only 1 is better params insted of param?

imthatguyhere: For consistency alone, yes.

DomT602: And performance yes.

life_server/Functions/Systems/fn_terrainSort.sqf
```diff
@@ -12,7 +12,7 @@ if !(params[["_dictionary",[],[[],0,""]]]) exitWith {};
 scopeName "main";
 
 {
-    if (worldName isEqualTo _x select 0) then {
+    if (worldName == _x select 0) then {
```
klmunday: How come the change to the `==` comparison?
`isEqualTo` is faster although case sensitive, I do not believe this has caused an issue as long as people are aware of this

koalas11: there was an error with IsEqualTo

DomT602: What error exactly? If people fill in the worldname correctly, its fine?

koalas11: Error in expression <e "main";

{
if (worldName isEqualTo _x select 0) then {
(_x select 1) breakOut >
  Error position: <select 0) then {
(_x select 1) breakOut >
  Error select: Type Bool, expected Array,String,Config entry
File \life_server\Functions\Systems\fn_terrainSort.sqf [TON_fnc_terrainSort], line 15

koalas11:
For now the changes on the FSM doesn't give any error on startup but i will do some more testing soon

koalas11:
ok remade the FSM files with arma tools

BlackFisch:
will this be updated in the near future? Otherwise I would create the PR again with updated base

DomT602:
With the SQL Custom update which affected the majority of the server sided files, including some optimisations, I'd agree with @BlackFisch on using the updated base if you'd like to continue with this.

DomT602:
Due to inactivity with this PR, it will now be closed.

PR #578, opened 2019-10-14
rework fn_useItem.sqf to use edible attribute

BlackFisch:
### Changes proposed in this pull request: 
- changing `fn_useItem.sqf` to automatically _'detect'_ whether an item is edible or not and let that be the default case (removing the necessity to manually add all edibles)
- remove old switch for hunger validation by a slim min/max command combo (performance)

- [x] I have tested my changes and corrected any errors found (actively used on our server for months now)


koalas11:
you can remove the [] before call and you can put together 
 
life_hunger = ((life_hunger + (M_CONFIG(getNumber,"VirtualItems",_item,"edible"))) max 5) min 100; //never below 5 or above 100
and if you change that i suggest testing it :P

BoGuu:
```
you can remove the [] before call and you can put together

life_hunger = ((life_hunger + (M_CONFIG(getNumber,"VirtualItems",_item,"edible"))) max 5) min 100; //never below 5 or above 100
and if you change that i suggest testing it :P```

All of which will make it less readable

**Comment by BlackFisch** [comment]
> you can remove the [] before call and you can put together
> 
> life_hunger = ((life_hunger + (M_CONFIG(getNumber,"VirtualItems",_item,"edible"))) max 5) min 100; //never below 5 or above 100
> and if you change that i suggest testing it :P

I don't think that such a long one-liner would be that great to understand, especially to beginners.
With the brackets - just sticking to project "guidelines" (at least that's how it's done almost everywhere...)

**Comment by Leigham** [comment]
`never below 5` wouldn't this mean the player would never starve ?

**Comment by koalas11** [comment]
Coding it's not always about readability.

**Comment by Leigham** [comment]
> Coding it's not always about readability.

maybe if your not the only one looking at it.

**Comment by koalas11** [comment]
yeah and it's too hard to read
Call something 
instead of
[] call something
tooo hard work for normal people

**Comment by Leigham** [comment]
its more of the putting it all together that hinders readability.

**Comment by BlackFisch** [comment]
> 
> 
> yeah and it's too hard to read
> Call something
> instead of
> [] call something
> tooo hard work for normal people

as i said.... that's about consistency....
also - talking of readability -> especially for new people, i think the concept of parameters is already hard enough to grasp. so explaining `[] call func` calls with no parameters and `[3,2] call func` calls with 3 and 2 as param is more easy to understand....

either way - adding or removing those brackets don't contribute in any way to better execution time or anything related, so that's just a thing about preferences and project style (which I'm just sticking to....)

File: Altis_Life.Altis/core/pmenu/fn_useItem.sqf [hunk 1]
```diff
@@ -6,10 +6,21 @@
...
 */
-private "_item";
 disableSerialization;
 if ((lbCurSel 2005) isEqualTo -1) exitWith {hint localize "STR_ISTR_SelectItemFirst";};
-_item = CONTROL_DATA(2005);
+private _item = CONTROL_DATA(2005);
+
+if !(M_CONFIG(getNumber,"VirtualItems",_item,"edible") isEqualTo -1) exitWith {
+    if ([false,_item,1] call life_fnc_handleInv) then {
+        private _val = M_CONFIG(getNumber,"VirtualItems",_item,"edible");
```
DomT602: Thanks for the PR @BlackFisch. Looks mostly good, but just for my sanity can we avoid doing `M_CONFIG(getNumber,"VirtualItems",_item,"edible")` twice, so just define the var before, then check if the var isEqualTo -1 etc

BlackFisch:
yeah, I think I somehow missed these parts... thanks @Jason2605 @DomT602

Jason2605:
Thanks for this @BlackFisch!

Issue #584, opened 2019-12-08
[Money Bug] Sell a House / Garage Multiple times


TaktischerSpeck:
### Expected behaviour
The House menu closes bevor the sell button can be pushed multiple times

### Actual behaviour
With a Speedclicker (https://www.speedautoclicker.net/) you can "speed click" the sell button and the script gonna be executed multiple times.

### Steps to reproduce the behaviour
1. Buy a Garage / House (i mostly tried Garages)
2. Have a Speedclicker (https://www.speedautoclicker.net/ <- is a basic one, you can sell the house about 3-4 times with the default "save" settings)
3. Open die House menu (windows button)
4. Hover over the "Sell" button and activate the speed clicker
5. Check your Bank...

Video: https://www.youtube.com/watch?v=Z-YyZhhVED8
----

Mission version: 5.0, with a lot of modifications but none wich should change the behaviour of this bug

----

I already fixed it on my Server and would make an Merge-Request when someone can confirm the Bug

akane777222:
Hey boguu it's still vulnerable to autoclick. What TaktischerSpeck mentioned is very effective and auto-clicking is completely blocked.

JulianKropp:
Thanks for this fix!!!

PR #585, opened 2019-12-08
Issue 584 [House sell Bug]

TaktischerSpeck:
https://github.com/AsYetUntitled/Framework/issues/584

Resolves #. 

- 

### DIFF
--- a/Altis_Life.Altis/core/configuration.sqf
+++ b/Altis_Life.Altis/core/configuration.sqf
@@ -43,6 +43,7 @@ life_canpay_bail = true;
 life_storagePlacing = scriptNull;
 life_hideoutBuildings = [];
 life_firstSpawn = true;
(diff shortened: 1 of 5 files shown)


BoGuu:
Reversing the order may address this, we'll see (9d4ee25d97a2754c6c43655658a00b5bc74c5125)

Vlieker:
This doesn't work because gettingBought will be one before someone succesfully sells a house

Why not just add a bool that gets set to true after someone buys a house, and check for that bool before you change it to true

Also, use a spawn with a sleep to assign false to the bool, so the script doesn't stay open untill that's done

BoGuu:
> This doesn't work because gettingBought will be one before someone succesfully sells a house

Exactly, it's checking that it's greater than 1 for the exit condition

PR #590, opened 2019-12-21
rework drinking system

BlackFisch:
- refactor drinks to have individual "stats" (not every drink resets thirst to 100, can be adjusted)
- introduced new parameter `drinkable` for Items (Config_vItems) to adjust the "stats". `-1` disabled, other values = thirst adjustment (like `edible` attribute)
- refactor `fn_useItem.sqf` to allow for items to influence hunger & thirst at the sam time (think of a soup or milk?)


BlackFisch:
@BoGuu thanks for editing, think i somehow missed that :D also merry christman

PR #596, opened 2019-12-31
Death log update

DomT602:
Resolves #. 

- Move from MPKilled EVH to entityKilled EVH on server. This will be apparent after SQL Custom update.


Jason2605:
Tested this and works well 👍

PR #603, opened 2020-01-10
Reveal objects rework

DomT602:
Resolves #. 

- Prevents revealObjects doing double, cleanup of some surrounding variables
- Also improves player tags settings code


Altis_Life.Altis/core/configuration.sqf
```diff
@@ -47,8 +47,6 @@ life_firstSpawn = true;
 //Settings
 life_settings_enableNewsBroadcast = profileNamespace getVariable ["life_enableNewsBroadcast", true];
 life_settings_enableSidechannel = profileNamespace getVariable ["life_enableSidechannel", true];
-life_settings_tagson = profileNamespace getVariable ["life_settings_tagson", true];
-life_settings_revealObjects = profileNamespace getVariable ["life_settings_revealObjects", true];
```
BoGuu: This seemed to bring more clarity than inlining?

Altis_Life.Altis/core/functions/fn_revealObjects.sqf
```diff
@@ -8,8 +8,6 @@
     
     Can be taxing on low-end systems or AMD CPU users.
 */
-if (!life_settings_revealObjects) exitWith {};
```
BoGuu: I like this check - I've no idea if it's ever possible to reach this point, but avoiding the next part is preferable, nothing wrong with sanity

Altis_Life.Altis/core/pmenu/fn_s_onCheckedChange.sqf
```diff
@@ -6,30 +6,30 @@
...
+];
+if (_option isEqualTo "") exitWith {};
 
-switch (_option) do {
+switch _option do {
     case "tags": {
         if (_state isEqualTo 1) then {
-            life_settings_tagson = true;
             profileNamespace setVariable ["life_settings_tagson",true];
+            LIFE_ID_PlayerTags = addMissionEventHandler ["EachFrame", life_fnc_playerTags];
```
BoGuu: Double check LIFE_ID_PlayerTags for these before adding/removing (sanity)?

Ditto below

Altis_Life.Altis/core/functions/fn_revealObjects.sqf
DomT602: This check is pointless because as we should remove the EachFrame handler, instead of having the handler always on, and using a variable to stop it on line 1

Altis_Life.Altis/core/configuration.sqf
DomT602: The variables were only used twice, which is pointless to store for when its used so rarely.

PR #627, opened 2020-03-14
Update fn_weaponShopMags to get all available Ammo

Blauschimmel:
The current version is a hotfix I suggested a while ago. As I just realized, bohemia changed several config entries, so I came up with a way to cover all "muzzles" (all underbarrel weapons) and as using the magazineWell entries, I'm now able to adress all crossover ammo from other mods etc. I'm about to test it asap.

Resolves #. 

- 

### DIFF
--- a/Altis_Life.Altis/core/shops/fn_weaponShopMags.sqf
+++ b/Altis_Life.Altis/core/shops/fn_weaponShopMags.sqf
@@ -9,13 +9,31 @@
 disableSerialization;
 
 if ((uiNamespace getVariable ["Weapon_Magazine",0]) isEqualTo 0) then {
-    private _weapon = lbData[38403,lbCurSel (38403)];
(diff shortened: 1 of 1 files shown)


Blauschimmel:
@DomT602 Whst do you mean by checking for blank and no usage of private? I didn't work on Altis Life for quite a while now so what am I missing out?

DomT602:
> @DomT602 What do you mean by checking for blank and no usage of private? I didn't work on Altis Life for quite a while now so what am I missing out?

Hi @Blauschimmel, if you look at what was there previously you can see the usage of private as a Keyword (negligble performance impact) compared to the old style array. We were also moving away from `count _muzzles > 0` as `!(_muzzles isEqualTo [])` is proven to be better.

BlackFisch:
For file styling - try to look for an editor config-plugin for your text editor. That way you don't have to bother about spaces instead of tabs and stuff like that, cause the repo includes an editor config file

PR #632, opened 2020-03-17
Typo in var inside private array

miguel93041:
Renamed "_digots" var which is non-existent to "_digits"


Turdant:
#576
https://github.com/AsYetUntitled/Framework/blob/master/.github/CONTRIBUTING.md#guidelines
Guideline 2 :)

DomT602:
@miguel93041 thanks for the PR. As this is a BIS function and our aim is to rewrite this for the future, this PR will be closed.

PR #635, opened 2020-03-19
returns rental vehicles

Turdant:
- Returns rental vehicles, so people don't leave them out and about as they can't return them to the garage. Doesn't put them in their garage, just deletes the vehicle.


life_hc/MySQL/Vehicles/fn_vehicleStore.sqf
```diff
@@ -71,6 +71,9 @@ if (_vInfo isEqualTo []) exitWith {
     [1,"STR_Garage_Store_NotPersistent",true] remoteExecCall ["life_fnc_broadcast",_ownerID];
     life_garage_store = false;
     _ownerID publicVariableClient "life_garage_store";
+    if (!isNil "_vehicle" && {!isNull _vehicle}) then {
```
DomT602: Surely the deleting of the vehicle should be found within this if statement: https://github.com/AsYetUntitled/Framework/blob/7569490dc2b5d68c966fa8a543b20933b6d21c38/life_server/Functions/Systems/fn_vehicleStore.sqf#L61

Turdant: It is? But you seem to be linking some 5.x version, and this is the master branch

https://github.com/AsYetUntitled/Framework/blob/e91e463eeba881fdd6a52c1b2a857cca5da32917/life_server/Functions/Systems/fn_vehicleStore.sqf#L69

Turdant: And this is also the `life_hc` file not the `life_server` file

https://github.com/AsYetUntitled/Framework/blob/7569490dc2b5d68c966fa8a543b20933b6d21c38/life_hc/MySQL/Vehicles/fn_vehicleStore.sqf#L65

Turdant:
Something like this @DomT602 ?

PR #647, opened 2020-04-18
Spikestrip rework

### LINKED ISSUE #212: Placing spikestrips - can only place one

DomT602:
Resolves #212.

- Customisation of minimum speeds for spike strips to work, and max number of spikes placed per cop
-  Optimised remainder of spike-related files, including a full rewrite of server handling to prevent unnecessary threads being spawned


life_hc/MySQL/Items/fn_spikeStrip.sqf
```diff
@@ -1,24 +1,47 @@
...
+    Adds spikestrip to hc-side array and if required - starts the monitoring of spikestrips.
+*/
+params [
+    ["_spikeStrip", objNull, [objNull]]
+];
+if (isNull _spikeStrip) exitWith {};
+
+hc_spikes pushBack _spikeStrip;
+
+if (hc_spikes isEqualTo []) then { //start monitoring spikestrips
```
BlackFisch, line 18: you sure about the empty array here? think you're missing a negation

life_hc/MySQL/Items/fn_spikeStrip.sqf
```diff
@@ -1,24 +1,47 @@
...
+                    deleteVehicle _x;
+                    _toDelete pushBack _forEachIndex;
+                };
+            };
+        } forEach hc_spikes;
+
+        reverse _toDelete;
+        {
+            hc_spikes deleteAt _x;
+        };
```
BlackFisch, line 43:
```suggestion
        } forEach _toDelete;
```

life_server/Functions/Systems/fn_spikeStrip.sqf
```diff
@@ -1,22 +1,45 @@
...
+    Adds spikestrip to server-side array and if required - starts the monitoring of spikestrips.
+*/
+params [
+    ["_spikeStrip", objNull, [objNull]]
+];
+if (isNull _spikeStrip) exitWith {};
+
+server_spikes pushBack _spikeStrip;
+
+if (server_spikes isEqualTo []) then { //start monitoring spikestrips
```
BlackFisch, line 16: same as with hc

life_server/Functions/Systems/fn_spikeStrip.sqf
```diff
@@ -1,22 +1,45 @@
...
+                    deleteVehicle _x;
+                    _toDelete pushBack _forEachIndex;
+                };
+            };
+        } forEach server_spikes;
+
+        reverse _toDelete;
+        {
+            server_spikes deleteAt _x;
+        };
```
BlackFisch, line 41:
```suggestion
        } forEach _toDelete;
```

life_hc/MySQL/Items/fn_spikeStrip.sqf
```diff
@@ -1,24 +1,47 @@
...
+                _toDelete pushBack _forEachIndex;
+            } else {
+                if (isNull _object) then {
+                    deleteVehicle _x;
+                    _toDelete pushBack _forEachIndex;
+                };
+            };
+        } forEach hc_spikes;
+
+        reverse _toDelete;
```
BlackFisch, line 40: why reverse? shouldn't make a difference

blackfisch (changes requested):

life_hc/MySQL/Items/fn_spikeStrip.sqf
DomT602: Yes, if the array is blank (no spikes are out), the loop should be started.

BlackFisch: You add a spikeStrip to the array 2 lines above. It will not be empty at this point, no matter what.

`count hc_spikes isEqualTo 1`

life_hc/MySQL/Items/fn_spikeStrip.sqf
DomT602: You have to delete backwards as if I delete index 0, the hc_spikes array has a new index and so forth, hence why deleting from it backwards never affects earlier indices

life_hc/MySQL/Items/fn_spikeStrip.sqf
```diff
@@ -1,24 +1,47 @@
...
+
+        {
+            private _pos = getPosATL _x;
+            (nearestObjects [_pos,["Car"],5]) params [["_nearVeh", objNull]];
+            if (alive _nearVeh && {abs (speed _nearVeh) > _minSpikeSpeed}) then {
+                [_nearVeh] remoteExecCall ["life_fnc_spikeStripEffect",_nearVeh];
+                deleteVehicle _x;
+                _toDelete pushBack _forEachIndex;
+            } else {
+                if (isNull _object) then {
```
BlackFisch, line 33: what is `_object`?

life_hc/MySQL/Items/fn_spikeStrip.sqf
BlackFisch: you got a point there :D

One more thought...
Is saving an array of indexes and iterating over them to `deleteAt` faster?

since you delete the objects, they become `objNull` and a simple `hc_spikes = hc_spikes - [objNull]` should do the same without iterating again

life_hc/MySQL/Items/fn_spikeStrip.sqf
```diff
@@ -30,7 +30,7 @@ if (hc_spikes isEqualTo []) then { //start monitoring spikestrips
                 deleteVehicle _x;
                 _toDelete pushBack _forEachIndex;
             } else {
-                if (isNull _object) then {
+                if (isNull _x) then {
                     deleteVehicle _x;
```
BlackFisch, line 34: if it's null... then why delete?

at this point this whole `else {...}` part is obsolete

PR #652, opened 2020-04-30
Container bugfix

### LINKED ISSUE #601: Containers variable return null object
1. 
2. 
3. 
4.

DomT602:
Resolves #601

- When player disconnects, sets the 'containers' variable on owned houses to nil


life_server/Functions/Housing/fn_houseCleanup.sqf
```diff
@@ -1,21 +1,26 @@
...
+
+{
+    _x params ["_pos"];
+    _pos = parseSimpleArray _pos;
+    {
+        deleteVehicle _x;
+    } forEach (nearestObjects[_pos,["Box_IND_Grenades_F","B_supplyCrate_F"],12]);
+
+    private _house = nearestObject [_pos, "House"];
+    if !(_house getVariable ["containers",[]] isEqualTo []) then {
```
BlackFisch, line 23: why even check for empty array? simply setting it to nil should be always ok

PR #653, opened 2020-04-30
Container bugfix

### LINKED ISSUE #601: Containers variable return null object

DomT602:
Resolves #601 (on master branch)

- When player disconnects, sets the 'containers' variable on owned houses to nil


cryptearth:
doesn'T work in current master branch commit a5a5821
`Error in expression <syncCall;

{
_x params ["_pos"];
_pos = parseSimpleArray _pos;
private _house = >
  Error position: <parseSimpleArray _pos;
private _house = >
  Error parsesimplearray: Type Array, expected String
File \life_server\Functions\Housing\fn_houseCleanup.sqf [TON_fnc_houseCleanup]..., line 16`

cryptearth:
error seems to be in AL.ini
the PR has this
`[selectHousesCleanup]
SQL1_1 = SELECT pos FROM houses WHERE pid = ?
SQL1_INPUTS = 1
OUTPUT = 1`
but as parseSimpleArray expects a string the output has to set to "1-STRING" like this:
`[selectHousesCleanup]
SQL1_1 = SELECT pos FROM houses WHERE pid = ?
SQL1_INPUTS = 1
OUTPUT = 1-STRING`
then this works and resolvs the dup bug

DomT602:
Hi @cryptearth thanks for the comments and getting this tested. Good spot on the error and a suggested fix, however an even better fix is just to remove the parseSimpleArray usage - as SQL Custom can give us an array as an array (rather then a string as we did before). This will be the best way forward.

cryptearth:
So, as I'm at my sisters wedding only using my notebook I had to quickly write my own PBO packer https://github.com/cryptearth/pbopacker - pulled latest master commit 99e05e8 - quickly patched in your latest edits - and uploaded to my server.
From as far as I tested it works the same as my idea with change the return type to string and perform the array parse - but as extdb already returns an array your removal of the parsing effectively does the same - so it works the same - no issues, no duping of creates / vItems - from my test ready to merge.

cryptearth:
Any reaso why this wasn't merged yet?

PR #658, opened 2020-05-02
New loadout approach

Casperento:
- fn_startLoadout.sqf refact. to use setUnitLoadout


Altis_Life.Altis/core/functions/fn_startLoadout.sqf
```diff
@@ -6,76 +6,23 @@
...
-    };
-};
-
-if !(_pVest isEqualTo []) then {
-    _pVest apply {
-        if (!(_x isEqualTo []) && {!((_x select 0) isEqualTo "") && {([(_x select 1)] call life_fnc_levelCheck)}}) then {
-            player addVest (_x select 0);
+private _level = call {
+    if (playerSide isEqualTo civilian) exitWith {
+        if (life_is_arredted) then {
```
DomT602, line 11:
```suggestion
        if (life_is_arrested) then {
```

Altis_Life.Altis/core/functions/fn_startLoadout.sqf
```diff
@@ -6,76 +6,23 @@
...
     };
-};
-
-if !(_pBackpack isEqualTo []) then {
-    _pBackpack apply {
-        if (!(_x isEqualTo []) && {!((_x select 0) isEqualTo "") && {([(_x select 1)] call life_fnc_levelCheck)}}) then {
-            player addBackpack (_x select 0);
-        };
+    if (playerSide isEqualTo west) exitWith {
+        str(FETCH_CONST(life_coplevel))
```
DomT602, line 18: instead of using str around each of the levels (for civ, cop and med), an easier solution could be to use format on line 25:
`player setUnitLoadout (missionConfigFile >> "Loadouts" >> str(playerSide) >> format["lvl_"_level];`

Altis_Life.Altis/config/Config_Loadouts.hpp
```diff
@@ -2,77 +2,118 @@
...
-            {"ItemCompass", ""},
-            {"ItemWatch", ""}
+    // COP
+    class WEST {
+        class lvl_1 : lvl_base {
+            uniformClass = "U_Rangemaster";
+            weapons[] = {"hgun_P07_snds_F"};
+            magazines[] = {"16Rnd_9x21_Mag","16Rnd_9x21_Mag","16Rnd_9x21_Mag","16Rnd_9x21_Mag","16Rnd_9x21_Mag","16Rnd_9x21_Mag"};
+        };
+        class lvl_2 : lvl_base {
```
DomT602, line 64: Its good to see a base class being used, however you could be using the first class to avoid repeating yourself below, especially when all loadouts have the same.
```suggestion
        class lvl_2 : lvl_1 {};
```

Casperento:
Everything tested and working as expected

Casperento:
Sounds good. About the civilian classes, i'm not sure yet how can we work it. I'm opened to suggestions..

Altis_Life.Altis/core/functions/fn_startLoadout.sqf
```diff
@@ -6,23 +6,44 @@
     Description:
     Loads a custom loadout on player when he got a new life
 */
+private _side = call {
+    if (playerSide isEqualTo civilian) exitWith {"civ"};
+    if (playerSide isEqualTo west) exitWith {"cop"};
+    if (playerSide isEqualTo independent) exitWith {"med"};
+};
+
 private _level = call {
```
DomT602, line 15: Bit of duplication with line 31 here I think. I think we could do it better by doing something like this:
`
if (_side isEqualTo "civ") then {
    if (life_is_arrested) then {
        //arrestedLoadout
    } else {
        //randomLoadout
    };
} else {
    //fetchRelevantLevel
    //setLoadout  
};`

Altis_Life.Altis/core/functions/fn_startLoadout.sqf
```diff
@@ -6,23 +6,44 @@
...
+    private _arr = [
+        "U_C_Poloshirt_blue",
+        "U_C_Poloshirt_burgundy",
+        "U_C_Poloshirt_stripped",
+        "U_C_Poloshirt_tricolour",
+        "U_C_Poloshirt_salmon",
+        "U_C_Poloshirt_redwhite",
+        "U_C_Commoner1_1"
+    ];
+	_civLoadout set [3, [(selectRandom _arr), []]];
```
DomT602, line 42: So this is a funny one, as you're setting the whole index of 3 as this info which means we could lose items here. We could be extra fancy and select into the array and set there.
```suggestion
(_civLoadout select 3) set [0,selectRandom _arr]
```

Altis_Life.Altis/core/functions/fn_startLoadout.sqf
```diff
@@ -6,23 +6,44 @@
...
-    if (playerSide isEqualTo independent) exitWith {
+    if (_side isEqualTo "med") exitWith {
         FETCH_CONST(life_mediclevel)
     };
 };
 
-player setUnitLoadout (missionConfigFile >> "Loadouts" >> str(playerSide) >> format["lvl_%1",_level]);
+if (_side isEqualTo "civ" && _level isEqualTo "random") then {
+    private _civLoadout = getUnitLoadout (missionConfigFile >> "Loadouts" >> "civ_level_random");
+    private _arr = [
```
DomT602, line 33: Use getArray to fetch the list of random clothing from Config_Loadouts - just to keep everything in the same place to avoid people editing this file unnecessarily.

Altis_Life.Altis/config/Config_Loadouts.hpp
```diff
@@ -17,72 +18,52 @@
...
+        magazines[] = {};
+    };
+    class civ_level_arrested { // Arrested player's loadout
+        uniformClass = "U_C_WorkerCoveralls";
+        backpack = "";
+        linkedItems[] = {};
+        weapons[] = {};
+        items[] = {};
+        magazines[] = {};
     };
```
DomT602, line 37:
```suggestion

civRandomClothing[] = {"string1", "string2};
```

Altis_Life.Altis/core/functions/fn_startLoadout.sqf
Casperento: I've tried that and it didn't work.

PR #672, opened 2020-05-23
fix hudupdate on healing with FirstAidKit

### LINKED ISSUE #666: hud not updated when healed with FirstAidKit
### Expected behaviour
hud update to reflect new damag set to 0.25

### Actual behaviour
"nothin"

1. damage yourself until damage gets more than 0.25 (= less than 75% health)
2. heal with FirstAidKit
3. damage is set to 0.25 (= 75% health) but no hud update

N/A

Mission version: master branch commit a5a5821

Launcher version: N/A

Game version: N/A

Branch: see above

This is realted to #663 but as FirstAidKit and its function is vanilla ArmA I guess to set the virtual damage to 0.25 is a hook somewhere - but I wasn't able to find it yet.
Fix would be similar to #664 as if when healing is done and damage value is set to 0.25 the hud gets updated.

cryptearth:
Resolves #666

- fix hud update after healing with FAK

note: player-interaction not yet tested


DomT602:
I'll predict the generic error as the code was put in unscheduled environment, when it needs scheduled (use spawn not call) and you'll need to remoteExecCall the life_fnc_hudUpdate if the player is not the unit that is healed.

cryptearth:
Thanks for the help - I'll try that next ...

cryptearth:
> [...] and you'll need to remoteExecCall the life_fnc_hudUpdate if the player is not the unit that is healed.

About that: Although I think I may have an idea what you mean (could be wrong tho) as far as I understand the BI wiki this shouldn't be needed. The BI wiki states that the HandleHeal EVH only gets triggered where the _injured unit (the one being healed) is local. So, it shouldn't matter if one heals self or is being healed by another player. It also shouldn't make any difference if a FAK or Medkit is used.
Please correct me if I'm wrong on this and suggest edits.

cryptearth:
Still only update hud on local when healing myself ...

cryptearth:
Ok, so I basicly just copied fn_jailSys about how to call a function remote on a given unit. This should finally fix the issue ...

cryptearth:
#688

PR #677, opened 2020-05-26
fix selling vehicles

### LINKED ISSUE #676: can not sell vehicle
### Expected behaviour
let the confirmation dialog come up as in #634

### Actual behaviour
fn_sellGarage errors out as lbCurSel returns -1 cause the dialog was closed too early in impound.hpp

try to sell a vehicle from a garage

not applicable

Mission version:
Launcher version:
Game version:
Branch: current master

I'm really sorry to write this again - but just another instance of "not tested before pushing public". In #634 impound.hpp was changed so that the dialog is closed before fn_sellGarage is called - but this causes lbCurSel 2802 to return -1 instead of the index. It's basically a 1-to-1 copy from the unimpound function directly above. And obviously this was not tested after implementing - otherwise @BlackFisch should had noticed that for what ever reason when clicking on sell the script fails with the STR_Global_NoSelection hint. A simple debug like
`_curIndex = lbCurSel 2802; diag_log format ["%1", _curIndex];`
right at the very top of fn_sellGarage show this error. This is not about "master branch is not recommended" but rather "devs just think and believe their changes do what they intent and push without any testing" - this is a serious process issue - not lack of coding skills in the SQF script language.

cryptearth:
Resolves #676

- as the dialog was closed too early lbCurSel in fn_sellGarage always returned -1 instead of cursor index

### DIFF
--- a/Altis_Life.Altis/dialog/impound.hpp
+++ b/Altis_Life.Altis/dialog/impound.hpp
@@ -76,7 +76,7 @@ class Life_impound_menu {
         class SellCar: Life_RscButtonMenu {
             idc = -1;
             text = "$STR_Global_Sell";
-            onButtonClick = "closeDialog 0; [] spawn life_fnc_sellGarage;";
+            onButtonClick = /*"closeDialog 0;*/ "[] spawn life_fnc_sellGarage;";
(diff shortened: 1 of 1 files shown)


cryptearth:
#688

PR #688, opened 2020-05-28
fixes for master branch

### LINKED ISSUE #683: cruel world: we can't fish anymore - manual fishing not working
### Expected behaviour
When killed a fish by SDAR swim near it, press WIN key, pick it up.

### Actual behaviour
Killing works - but the rest doesn't > ends up in log spam.

1. kill a fish
2. press WIN when near it
3. enjoy log spam

```
13:31:00 Error in expression <hName = localize _fishName;


if ([true,_type,1] call life_fnc_handleInv) then {>
13:31:00   Error position: <_type,1] call life_fnc_handleInv) then {>
13:31:00   Error Undefined variable in expression: _type
13:31:00 File core\actions\fn_catchFish.sqf [life_fnc_catchFish]..., line 93
```

Mission version: 6.0.0 master commit ee0fbc3

Launcher version: 1.5.146373

Game version: 1.98.146373

cryptearth:
Resolves #683 #676 #691

- fork rebase

- [X] I have tested my changes and corrected any errors found


BoGuu:
re. HandleHeal - I'm correct in thinking you can cancel the healing right? 

If that's the case, I'm not at all a fan of HandleHeal - it seems pretty useless.

cryptearth:
Well, the main issue I try to resolve with it is that when a heal item (FirstAidKit, Medkit) is used on self or by other player in the field although the healing itself is done by the engine the hud is not updated. I've done it for the healing at the hospital NPC and hence I'd like to also get it working for the healing items. Unfortunately I wasn't able to find any other event handler triggered on healing. So, if you have any suggestions for this problem please tell me. Otherwise just not fix this at all isn't really satisfying me.

BoGuu:
Updating would be the default behavior on the v5 branch - but a damage check was removed in a tidy up. The easiest solution to this is to utilise `life_fnc_survival`, it does periodic state checking as is, one would just need to re-add a damage check between iterations

cryptearth:
I don't see such a change would fix the base issue as the hud gets updated everytime either food/water ticks or if the player consumes some item that change those values. So in other words: Getting healed (either self or by someone else) and using a water bottle also end up in refreshing the hud - but the issue still remains that the healing isn't caught so the "health" (or rather: damage) indicator doesn't get updated right away but only when a change on one of the other two triggers the hud update.
On the other hand, when damage is inflicted the hud gets update straight away. So at least it's some handler triggers on damage infliction. I just want to get the same working when healing happens.

From as far as I now understand it's this: It seems that the event handler for the "HandleHeal" event always gets triggered on the unit using the FAK, no matter if this unit heals itself or another player.
As the BI wiki itself shows there's an easy to wait until the healing itself is done and then continue with triggering something else. My next idea now is to relay the completion of the healing to the server and let it relay it back to the unit that's being healed. I tried to copy over from what I've seen in the jail system as it seems to do exactly what I want to achieve, but this would also require not just adding a simple event handler in the client mission but also some server code. I don't know if this would be worth a valuable modification to fix the hud not updating on healing.

It's one of those "This has to work somehow ..." type of things where the "correct" solution is so near but yet so far. The overall bad design of the engine, its scripting and the documentation doesn't help much in solve this issue.

cryptearth:
removed any changes for #666 as it's not possible to do

PR #694, opened 2020-06-01
blastingChargeFix

### LINKED ISSUE #690: fed bank need some rework
branch: master
commit: 86f766e

Sorry for not follow issue template this time, but it's to follow for this topic.

So I continued testing with my friends and we tested robbing the bank (I set required cops to 1 for testing). As I also wanted to test the bomb defusal one of us planted the bomb, I defused it, and the bomb was replanted. While the timer still had about 2m to go suddenly the explosion triggered and the safe was opened.
After digging through the involved files I suspect this to be the issue:
life_server/Functions/Systems/fn_handleBlastingCharge.sqf
```
private ["_bomb","_time"];
_time = time + (5 * 60);
waitUntil{(round(_time - time) < 1)};
sleep 0.9;
if (!(fed_bank getVariable["chargeplaced",false])) exitWith {};

_bomb = "Bo_GBU12_LGB_MI10" createVehicle [getPosATL fed_bank select 0, getPosATL fed_bank select 1, (getPosATL fed_bank select 2)+0.5];
fed_bank setVariable ["chargeplaced",false,true];
fed_bank setVariable ["safe_open",true,true];
```
The issue is this:
The timer gets started when the first bomb is planted. It ticks down for the full 5m no matter what else happen on the server. After the 5m is done then it's checked again if in that very moment any bomb is planted, no matter if it's the same or another as the original was defused. If at least some bomb is planted when the first timer hits 0 the explosion is triggered, any other bomb is "disarmed" (so their timers just run out without anything else happening) and the vault is opened.
We didn't tested it, but I suspect that with good timing and a bit of the help of at least one of the cops the bank maybe can be looted twice within that 5m timer interval; although it would require some precise timing and planing. Maybe subject for another test.

How can this be fixed: I'm not sure how well the ArmA3 engine can be manipulated while waiting for something to happen in a not-so-resource-impacting-style, but if possible the timer should be somehow "aborted" when the bomb is defused. Maybe extending the condition to something like this:
`waitUntil{(round(_time - time) < 1)||(fed_bank getVariable["chargeplaced",false])};`

Turdant:
Resolves #690 

- Checks whether the bomb gets defused on each frame, and if the bomb gets defused it exits the `waitUntil` loop and thus checks whether the bomb is placed (which it is not), so it exits the function.

EDIT: Typo


life_server/Functions/Systems/fn_handleBlastingCharge.sqf
```diff
@@ -4,12 +4,11 @@
     File Name: fn_handleBlastingCharge.sqf
     Information: Nothing really special, just handles the fed explosion over the server so if the client who place the charge logs out it still blows up.
 */
-private ["_bomb","_time"];
-_time = time + (5 * 60);
-waitUntil{(round(_time - time) < 1)};
+private _time = time + (5 * 60);
+waitUntil{(round(_time - time) < 1) || !(fed_bank getVariable["chargeplaced",false])};
 sleep 0.9;
```
DomT602, line 9: Do we need this sleep?

life_server/Functions/Systems/fn_handleBlastingCharge.sqf
```diff
@@ -4,12 +4,11 @@
...
 if (!(fed_bank getVariable["chargeplaced",false])) exitWith {};
```
DomT602, line 10: Might as well remove the extra brackets I guess

life_server/Functions/Systems/fn_handleBlastingCharge.sqf
```diff
@@ -4,12 +4,11 @@
     File Name: fn_handleBlastingCharge.sqf
     Information: Nothing really special, just handles the fed explosion over the server so if the client who place the charge logs out it still blows up.
 */
-private ["_bomb","_time"];
-_time = time + (5 * 60);
-waitUntil{(round(_time - time) < 1)};
+private _time = time + (5 * 60);
```
DomT602, line 7: Thinking about it - we can make this PR a bit better with giving a configurable bomb timer? It'd need edits in here, fn_demoChargeTimer and Config_Master.

life_server/Functions/Systems/fn_handleBlastingCharge.sqf
```diff
@@ -4,12 +4,12 @@
     File Name: fn_handleBlastingCharge.sqf
     Information: Nothing really special, just handles the fed explosion over the server so if the client who place the charge logs out it still blows up.
 */
-private ["_bomb","_time"];
-_time = time + (5 * 60);
-waitUntil{(round(_time - time) < 1)};
-sleep 0.9;
-if (!(fed_bank getVariable["chargeplaced",false])) exitWith {};
+private _time = time + LIFE_SETTINGS(getNumber,"fed_chargeTime") * 60;
```
shazoo-o, line 7: LIFE_SETTINGS require include of script_macros.hpp

DomT602:
@shazoo-o good spot - have you by chance tested it with macros in the file?

shazoo-o:
> @shazoo-o good spot - have you by chance tested it with macros in the file?

I noticed this problem as I had merged this in my repo and tried to start the server - which failed with an error message around the LIFE_SETTINGS macro. I then just added the include line and it worked.
I then also tested the base issue if the timer really is stopped when the bomb gets defused which, at least in my test, also worked as intended.

PR #695, opened 2020-06-01
Refactor configs to use preprocessed Strings

BlackFisch:
As the [Bohemia Wiki entry for Stringtable.xml](https://community.bistudio.com/wiki/Stringtable.xml#Config) indicates, it is possible to add Strings in Configs with `$` as suffix, which preprocesses them into the config itself making a call like `localize _title` unnecessary.
We already use this in some places in the Framework, like for the `Remove Gear` options in the clothing shops, and I see no real benefit in just doing it everywhere else too.

As a side effect, this also allows for people entering plain-text displaynames, if the don't want or need localization on their server (or are just lazy folks like I am).

#### ToDo
- [X] Clothing Shop titles
    - [x] tested

- [x] Licenses
    - [x] tested

- [x] Wanted Crimes
    - [x] tested

- [x] Process texts
    - [x] tested

- [x] Virtual Shop titles
    - [x] tested

- [x] Virtual Item names
    - [x] tested

- [x] Weapon Shop names
    - [x] add localization (why are they even plain text?)
    - [x] tested

#### Final check


BlackFisch:
tested everything I could, only things not actively checked is the multiplayer stuff - which should still be fine since I only removed `localize`

- Cop searching player (Altis_Life.Altis/core/cop/fn_copSearch.sqf & Altis_Life.Altis/core/cop/fn_licenseCheck.sqf)
- transferring items between players (Altis_Life.Altis/core/functions/fn_giveDiff.sqf, Altis_Life.Altis/core/functions/fn_receiveItem.sqf & Altis_Life.Altis/core/pmenu/fn_giveItem.sqf)
- clist of crimes on player (Altis_Life.Altis/core/pmenu/fn_wantedInfo.sqf)

WolfSkin1337:
with this changes, the license dont shown up at the Virtual-Menu, but i have buy it. and i can Process the item

PR #704, opened 2020-06-03
[master] fix variable shadowing in fn_fuelSupply

shazoo-o:
- block-local _fuelFeedState in line 48 shadows global _fuelFeedState in line 40
to prevent shadowing block-local variable was renamed
- additional else-block to set outer _fuelFeedState when fuel station wasn't fully refilled

### DIFF
--- a/Altis_Life.Altis/core/vehicle/fn_fuelSupply.sqf
+++ b/Altis_Life.Altis/core/vehicle/fn_fuelSupply.sqf
@@ -45,12 +45,14 @@ private _random = floor((random 11000) + 1500);
         _x setVariable ["fuelTank",[_random,time],true];
         _fuelFeedState = _random;
     } else {
-        _stationTank params ["_fuelFeedState","_fuelFeedTime"];
-        if (_fuelFeedState isEqualTo 0) then {
+        _stationTank params ["_pFuelFeedState","_fuelFeedTime"];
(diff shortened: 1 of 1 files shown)


BlackFisch:
This isn't a real issue, since it's declared inside the if andd thus it's another scope - the other variable isn't tempered with by that
For readability though - I approve :D

shazoo-o:
I've tested it quite well.
The issue is this: https://en.wikipedia.org/wiki/Variable_shadowing

How it fails:

1. On the first run line 43 returns the default empty array as the member fuelTank does not yet exists for this fuel pump. Then line 46 sets _fuelFeedState to the random value so line 59 is passed this first time.
2. On every other run later fuelTank is already defined so the else block is executed. The params[] then creates a new additional _fuelFeedSate which shadows the one declared in line 40. As the outer one in line 40 then is never changed again the if() on line 59 always triggers and cause exitWith().
3. The additional else block on line 54 is required for the same reason: If the fuel station isn't fully refilled the _fuelFeedState of the current line 48 shadows the one from line 40 so the current fuel level never gets to line 59 which again result in exitWith() is always triggered.

Sorry I couldn't come up with a better new name, but it shouldn't matter much.

BlackFisch:
Oh, I see the problem :D didn't recognize this on the phone ^^ Good spot!

shazoo-o:
fix for 5.x.x in #705

PR #710, opened 2020-06-04
paycheckRework

Turdant:
- Changing the paychecks to a rank based approach

### DIFF
--- a/Altis_Life.Altis/config/Config_Master.hpp
+++ b/Altis_Life.Altis/config/Config_Master.hpp
@@ -125,9 +125,9 @@ class Life_Settings {
     bank_civ = 3000; //Amount of cash in bank for new civillians
     bank_med = 6500; //Amount of cash in bank for new medics
 
-    paycheck_cop = 500; //Payment for cops
+    paycheck_cop = [500,550,600,650,700,750,800,850]; //Payment for cops, increases with rank [rank 0, rank 1, rank 2, etc.]
     paycheck_civ = 350; //Payment for civillians
-    paycheck_med = 450; //Payment for medics
+    paycheck_med = [450,500,550,600,650,700]; //Payment for medics, increases with rank [rank 0, rank 1, rank 2, etc.]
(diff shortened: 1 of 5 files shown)


Altis_Life.Altis/config/Config_Master.hpp
```diff
@@ -126,9 +126,9 @@ class Life_Settings {
     bank_civ = 3000; //Amount of cash in bank for new civillians
     bank_med = 6500; //Amount of cash in bank for new medics
 
-    paycheck_cop = 500; //Payment for cops
+    paycheck_cop = [500,550,600,650,700,750,800,850]; //Payment for cops, increases with rank [rank 0, rank 1, rank 2, etc.]
     paycheck_civ = 350; //Payment for civillians
-    paycheck_med = 450; //Payment for medics
+    paycheck_med = [450,500,550,600,650,700]; //Payment for medics, increases with rank [rank 0, rank 1, rank 2, etc.]
```
shazoo-o, line 131: Well, another one of those "the engine gut us again" things.
I noticed that my cops got $0 as paycheck and suddenly their bank accounts were also at $0. This reminded me of the pickupMoney bug where CASH got nil at some point - and sure it's the same issue here. I tested quite a bit, but in the end it came down to that for what ever reason paycheck_cop and paycheck_med are neither an array nor a number but rather a text (I tested this by unrolling the LIFE_SETTINGS macro and tested with isArray, isNumber and isText to see what it really was I got returned.
Then I looked at the other arrays in the config and spotted the issue: In the config an array has to be defined with the brackets after the variable name and the values have to be enclosed in curly braces:
```
paycheck_cop[] = { 500, 550, 600, 650, 700, 750, 800, 850 };
paycheck_med[] = { 450, 500, 550, 600, 650, 700 };
```

Turdant: Thank you! Should be fixed in the latest commit.

PR #721, opened 2020-06-12
Paycheck during death fix

### LINKED ISSUE #720: logical error - death and paycheck
### Expected behaviour
when dead not getting a paycheck but instead the text "you missed a paycheck"

### Actual behaviour
still getting a paycheck as by how "death" is implemented and checked

should be obvious - but k
1. die / get killed
2. wait up to (paycheck time here) before respawn
3. see you still get your paycheck even when dead

not applicable

Mission version: 6.0.0

Launcher version: 1.5.146373

Game version: 1.98.146373

Ok - so a bit more although I guess all active devs know the issue:
The way "death" is implement is to copy the player as a corpse and respawn at the spawn island. The issue: "isAlive player" returns true as the player isn't dead but only a fixed camera is set on the corpse.
Possible solution: Set a flag on the player (isn't such already done?) and check it instead of isAlive to check if the player currently is considered "dead" or "alive".

that-edward:
Resolves #720

- Change the condition in the if statement

### DIFF
--- a/Altis_Life.Altis/core/fn_survival.sqf
+++ b/Altis_Life.Altis/core/fn_survival.sqf
@@ -47,7 +47,7 @@ _fnc_water = {
 };
 
 private _fnc_paycheck = {
-    if (alive player) then {
+    if (life_is_alive) then {
         private _paycheck = call life_paycheck;
         if (player distance (getMarkerPos "fed_reserve") < 120 && playerSide isEqualTo west) then {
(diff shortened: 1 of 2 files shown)


DomT602:
So, looking into this issue a bit more, this PR looks like it'd work but looking into the `life_is_alive` variable further it seems a bit more work may be needed. Once the player dies it is set to false, as expected. If revived, it is reset to true as expected. However, the issue lies with when you respawn - the variable does not get set to true.

PR #734, opened 2020-07-29
Update fn_clearVehicleAmmo.sqf

FlorianGH:
- Changed If to switch

It's just a little thing but it needs to be done.

### DIFF
--- a/Altis_Life.Altis/core/functions/fn_clearVehicleAmmo.sqf
+++ b/Altis_Life.Altis/core/functions/fn_clearVehicleAmmo.sqf
@@ -14,24 +14,23 @@ _vehicle = [_this,0,objNull,[objNull]] call BIS_fnc_param;
 if (isNull _vehicle) exitWith {};
 _veh = typeOf _vehicle;
 
-if (_veh isEqualTo "B_Boat_Armed_01_minigun_F") then {
-    _vehicle removeMagazinesTurret ["200Rnd_40mm_G_belt",[0]];
-};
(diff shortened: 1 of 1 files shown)


BlackFisch:
moving the `clearWeaponCargoGlobal _vehicle;` to the top and using exitWith instead of then (or a switch) would actually be the best solution considering performance

Altis_Life.Altis/core/functions/fn_clearVehicleAmmo.sqf
```diff
@@ -12,29 +12,30 @@
...
-    _vehicle removeMagazinesTurret ["250Rnd_30mm_APDS_shells",[0]];
-    _vehicle removeMagazinesTurret ["8Rnd_LG_scalpel",[0]];
-    _vehicle removeMagazinesTurret ["38Rnd_80mm_rockets",[0]];
-};
-
-if (_veh isEqualTo "B_Heli_Transport_01_F") then {
-    _vehicle removeMagazinesTurret ["2000Rnd_65x39_Belt_Tracer_Red",[1]];
-    _vehicle removeMagazinesTurret ["2000Rnd_65x39_Belt_Tracer_Red",[2]];
-};
+private ["_vehicle","_veh"];
```
DomT602, line 15: We've managed to duplicate these 3 lines twice. It'd be best to remove the BIS_fnc_param and the private ARRAY and replace with params (obviously removing the duplication).

PR #743, opened 2021-01-25
Patch Inject Batiment hack script - 

ghost:
No-Inject Batiment hack script

Resolves #. 

- 


Jonastex (approved):
Why not, it blocks the creation of a vehicle that is not defined in the config, but it doesn't completely solve the issue

mavrickthiry:
I think me most effeciant thing to block that, is to setup a security-token.

The server generate a current "session" token for the player and all events, triggers is securised by this security-token client checked by the server.

